PR #7, opened 2018-04-05
Update the discount badge

vitoriaHeliane:
Update the discount badge and fix the discount badge mock.

Update the calculate discount tax method to receive the listPrice and sellingPrice as params instead of get them from the props. And update the label default value to an empty string instead of 'OFF'. Also, add a negative signal before the percent if the label isn't given. In the mock, add a json object to get the discount badge values from it.

#### Types of changes
- [ X ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/react/DiscountBadge.js
+++ b/react/DiscountBadge.js
@@ -8,18 +8,20 @@ import PropTypes from 'prop-types'
 */
 class DiscountBadge extends Component {
   /**
-  * This method calculates the discount tax by the selling and list prices.
+  * This method calculates the discount tax.
+  * @param {number} listPrice - The product's price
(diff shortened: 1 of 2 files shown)


react/DiscountBadge.js
```diff
@@ -8,18 +8,21 @@ import PropTypes from 'prop-types'
 */
 class DiscountBadge extends Component {
   /**
-  * This method calculates the discount tax by the selling and list prices.
+  * This method calculates the discount tax.
```
brunojdo: remove the jsdoc

vitoriaHeliane: Feito!

PR #14, opened 2018-04-13
Add initial tests to the Price component

estacioneto:

Add a test environment and some initial tests to begin to test all the components

There were no tests before.

```sh
# In react folder, run:
$ yarn test
# or...
$ npm test
```

image

#### Types of changes
- [x] New feature (non-breaking change which adds functionality)


react/__tests__/Price.test.js
```diff
@@ -0,0 +1,76 @@
...
+      maximumFractionDigits: 2,
+    }
+
+    productPropsMock = {
+      listPrice: 200,
+      sellingPrice: 170,
+      installments: 3,
+      installmentPrice: 50,
+    }
+    configuration = {}
```
lucasecdb: Don't rely on global variables to pass props to the render function, as it makes the code harder to read. A better alternative would be to just add an argument to the `renderComponent` function like this
```jsx
const renderComponent = (props = {}) => mountWithIntl(...)
```
Also, you don't need to redefine this function before each test case, you can put the implementation outside the `beforeEach` function callback.

react/__tests__/helpers/intlHelper.js
```diff
@@ -0,0 +1,30 @@
...
+  const childContextTypes = {
+    intl: intlShape,
+    culture: PropTypes.object,
+  }
+
+  return { context, childContextTypes, locale: DEFAULT_LOCALE }
+}
+
+export function getIntlInstance(props = { locale: DEFAULT_LOCALE }, context = {}) {
+  props.messages = props.messages || getMessages(props.locale)
```
lucasecdb: Don't alter function parameters. Create a new variable or use the object destructuring on usage, e.g
```js
const intlProvider = new IntlProvider({ message: props.message || getMessages(props.locale), ...props }, context)
```

react/__tests__/Price.test.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  it('should be mounted and not break', () => {
+    const { component } = renderComponent()
+    expect(component).toMatchSnapshot()
+  })
+
+  it('should not show the list price if prop showListPrice is false', () => {
+    const props = Object.assign({}, productPropsMock, defaultConfiguration, { showListPrice: false })
```
lucasecdb: Put this inside the `renderComponent` function.

react/__tests__/Price.test.js
```diff
@@ -0,0 +1,85 @@
...
+    const { component, context, intl } = renderComponent(props)
+    const currencyOptions = getCurrencyOptions(context)
+
+    expect(component.prop('showListPrice')).toBe(false)
```
lucasecdb: This line is redundant, there's no need to test props

react/__tests__/Price.test.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  it('should be mounted and not break', () => {
+    const { component } = renderComponent()
+    expect(component).toMatchSnapshot()
+  })
+
+  it('should not show the list price if prop showListPrice is false', () => {
+    const { component, context, intl } = renderComponent(Object.assign({}, productPropsMock, defaultConfiguration, { showListPrice: false }))
```
lucasecdb: Move the `Object.assign` call to *inside* the `renderComponent` function

react/__tests__/Price.test.js
```diff
@@ -0,0 +1,84 @@
...
+      style: 'currency',
+      currency: context.culture.currency,
+      minimumFractionDigits: 2,
+      maximumFractionDigits: 2,
+    }
+  }
+
+  it('should be mounted and not break', () => {
+    const { component } = renderComponent()
+    expect(component).toMatchSnapshot()
```
klzns: Pode usar uma outro tipo de assertiva pra esse teste. Acho snapshot pra esse caso overkill.

react/__tests__/Price.test.js
```diff
@@ -0,0 +1,84 @@
...
+  })
+
+  it('should not show the list price if prop showListPrice is false', () => {
+    const { component, context, intl } = renderComponent({ showListPrice: false })
+    const currencyOptions = getCurrencyOptions(context)
+
+    expect(component.contains(intl.formatMessage({ id: 'pricing.from' }))).toBe(false)
+    expect(component.contains(intl.formatMessage({ id: 'pricing.to' }))).toBe(true)
+    expect(component.contains(intl.formatNumber(productPropsMock.sellingPrice, currencyOptions))).toBe(true)
```
klzns: O que acha desses valores ficarem em uma linha a parte?

Ex:

```js
const sellingPrice = intl.formatNumber(productPropsMock.sellingPrice, currencyOptions))

expect(component.contains(sellingPrice)).toBe(true)
```

react/__tests__/Price.test.js
estacioneto: Nesse teste, teoricamente não precisaria de assertiva, pois se o componente quebrar, o teste quebra. Alguma sugestão de assertiva, @klzns ?

klzns: Acho que se a variavel `component` tiver definida, ele nao quebrou.

De qualquer forma quebra ele colocando um throw no render e ve se o teste de fato pega.

estacioneto: Entendi. Vou modificar

estacioneto: E sim, os testes falham quando uma exceção é lançada no render

PR #24, opened 2018-04-24
Fix product summary css, hover on mobile and default schema props

### LINKED ISSUE #15: Hide listPrice was equaled to sellingPrice
## Expected Behavior
If `listPrice` was equaled to `sellingPrice` the Price component must hide them. 

## Current Behavior
Isn't been hide.

screen shot 2018-04-18 at 13 52 03

### LINKED ISSUE #18: Hover don't exist in mobile screens
When show button on hover is active on the desktop, on mobile, we need to make the button stay show.

### LINKED ISSUE #23: Set default values in the ProductSummary Schema
The schema properties that have default value equal to `true`, must have a default value in the `ProductSummary.schema.properties`. Otherwise the ComponentEditor will display the value as  `false` 

```javascript
ProductSummary.schema = {
  title: 'Product Summary',
  description: 'The product summary showing the main product informations',
  type: 'object',
  properties: {
    showListPrice: {
      type: 'boolean',
      title: "Show product's list price",
      default: true // <- missing
    },
    ...
```

estacioneto:

1. Fix the Product Summary CSS (details on screenshots)
2. Fix #15 (Hide list price if equals selling price)
3. Fix #18 (Hover doesn't exist on mobile and tablets)
4. Fix #23 (Default Schema props)

Fix Summary issues and it's styling

[Access workspace](https://estacio--storecomponents.myvtex.com/) and see the styling and the Schema props. Access from a smartphone or tablet to see the buy button if it's not hidden completely.

image

Summary specification: 
image

#### Types of changes


react/ProductName.js
```diff
@@ -10,7 +10,12 @@ class ProductName extends Component {
 
     return (
       <div className="vtex-product-name overflow-hidden">
-        {referenceCode} {name} {skuName} {brandName && `(${brandName})`}
+        <div className="f5 truncate">
+          {referenceCode} {name}
```
brunojdo: Title: productName & branName (Optional)
Subtitle: SKU Name

estacioneto: Retirei o ReferenceCode então.

react/ProductSummary.js
```diff
@@ -46,14 +47,14 @@ class ProductSummary extends Component {
       badgeText,
       buyButtonText,
       hideBuyButton,
-      showButtonOnHover,
     } = this.props
 
     const product = this.props.product || createProduct()
+    const showButtonOnHover = this.props.showButtonOnHover && !isMobile
```
leticiawanderley: I don't think it is necessary to add a dependency for this. You can try to use the media query classes, for example: `this.props.showButtonOnHover ? 'db' : 'db dn-ns'`

react/Price.js
```diff
@@ -67,9 +71,10 @@ class Price extends Component {
                   }}
                 />
               ) : (
-                <span>{installments}X {formattedInstallmentPrice}</span>
-              )
-              }
+                <span>
+                  {installments}X {formattedInstallmentPrice}
```
leticiawanderley: Try to use the unicode value for `times` https://stackoverflow.com/questions/12009545/what-is-the-proper-html-entity-for-the-x-in-a-dimension

react/ProductSummary.js
estacioneto: The thing is that tablets screen sizes are not that small. That's why I wanted to do a mobile detection. Should I ignore the tablets fact?

react/Price.js
estacioneto: I made the change, but what about the locales messages? They will be different

leticiawanderley: You can pass the span element containing the Unicode as a parameter to the locale text.

react/ProductSummary.js
leticiawanderley: I still think this is an overkill but have no better solution, so leave it the way it is.

PR #60, opened 2018-11-19
Fix SVG warnings

rerissondaniel:

Replace the unit in which the width and height of content loaders were defined.

The height and width props of the react-content-loader were being defined using percentage but as these values are used to define the viewBox of the content loader warnings were shown. This PR replace these percentage values with user units, adequate to each content loader.

[Access this workspace](https://svgerrorpd--storecomponents.myvtex.com/iphone-xs/p), open the console. Select slow 3g in the network tab so you can see the content loaders.

#### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/react/index.js
+++ b/react/index.js
@@ -147,6 +147,7 @@ class ProductSummary extends Component {
       }}
       width={100}
       height={56}
+      preserveAspectRatio="xMinYMin meet"
     >
       <rect width="100%" height="100%" />
     </ContentLoader>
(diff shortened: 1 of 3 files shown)


react/index.js
```diff
@@ -140,8 +140,8 @@ class ProductSummary extends Component {
         width: '100%',
         height: '100%',
       }}
-      height="100%"
-      width="100%"
+      width="100"
+      height="56"
```
luciannojunior: Shouldn't there be a property for maintaning proportion in smaller screens? I remember another prop in your others PR's

rerissondaniel: Added

PR #62, opened 2018-11-27
Fix buy button not render properly

JuanAlmeida12:
Fix buy button not render properly

Buy button render only on hover.

[Access this workspace](https://fixbuybutton--storecomponents.myvtex.com/)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Fixed
-- Fix how buy button is rendered
+- Now, buy button doesn't move the product summary box when is hidden.
 
 ## [1.6.0] - 2018-11-26
 ### Added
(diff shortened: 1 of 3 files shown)


react/index.js
```diff
@@ -223,6 +223,12 @@ class ProductSummary extends Component {
       'bb b--muted-4 ma2': showBorders
     })
 
+    const getBuyButtonVisibility = () => {
+      return !((showButtonOnHover || showBuyButton) && showBuyButton) ? 'hidden' : 'visible'
```
luciannojunior: `showButtonOnHover` does not matter on this clause. The visibility only depends on `showBuyButton` in this case. If `true`, 'visible, if not `hidden`.

What's the expected behavior?

react/index.js
```diff
@@ -223,6 +223,12 @@ class ProductSummary extends Component {
...
+    }
+
+    const style = { visibility: getBuyButtonVisibility() }
```
luciannojunior: Use `tachyons`'s classes instead. `classname` could do it!

JuanAlmeida12: `tachyons` don't support change the `visibility` attribute.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fix how buy button is rendered
```
brunojdo: Change to: "Now, buy button doesn't move the product summary box when is hidden."

PR #63, opened 2018-11-27
Feature/update store components 3.x

Klynger:
Update the usage of components from `store-components`

`store-components` has breaking change and there is the need to update the usage of it's components

[Workspace](https://next--storecomponents.myvtex.com/)

captura de tela de 2018-11-27 16-14-15

*This PR rely in the merge of https://github.com/vtex-apps/store-components/pull/197*

#### Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Update `store-components` to `3.x`.
 
 ## [1.6.2] - 2018-11-27
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Changed
-- Update `store-components` to `3.x`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Update `store-components` to `3.x`.
```
brunojdo: Add Design Tokens on Product Summary.

react/index.js
```diff
@@ -100,7 +100,7 @@ class ProductSummary extends Component {
     return path(['sku', 'seller', 'commertialOffer'], this.props.product)
   }
 
-  renderImage = () => {
+  get renderImage() {
```
lucasecdb: why use the `get`? this is more like a method than a computed property, and makes more sense to call it as a ordinary method.

rerissondaniel: I think that if the get will be used would be better to remove the 'render' prefix. I think the get could be used since these methods dont change any fields

luciannojunior: `get` is fine, as @brunojdo said

Klynger: The idea of using getters is because is lazy. This fact wouldn't make it better use getters instead of ordinary method? @lucasecdb

lucasecdb: agree with @rerissondaniel

PR #68, opened 2018-12-05
[WIP]Use product image according to client device

rerissondaniel:

Use resized images depending on the client device. 

The image that was used had always the same size, it could make its loading slow, depending on the network. 

[workspace](https://imagesizing--storecomponents.myvtex.com/)
Check that the product image size is different in mobile and desktop.

#### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


react/components/Image.js
```diff
@@ -1,15 +1,39 @@
...
+  resizeUrl = src => {
+    //Check of src is already in a size defined by the client
+    const resized = /[1-9]+-(:?([1-9]+)|(auto))/
+    if (resized.exec(src)) {
+      return src
+    }
+
+    const { runtime: { hints: { mobile } } } = this.props
+
+    const imageId = src.slice(src.indexOf('/ids/')).split('/')[2]
```
luciannojunior: Duplicated code from https://github.com/vtex-apps/product-details/blob/master/react/utils/generateUrl.js#L22.

@brunojdo do you agree we should have a HOC in `store-components` that executes this behavior?

react/components/Image.js
```diff
@@ -1,15 +1,39 @@
...
-    <div
-      className={classNames(className, 'vtex-aspect-ratio w-100')}
-      style={{ backgroundImage: `url(${src})` }}
-    />
-  )
+class Image extends Component {
+
+  resizeUrl = src => {
+    //Check of src is already in a size defined by the client
+    const resized = /[1-9]+-(:?([1-9]+)|(auto))/
```
luciannojunior: `const isWithMeasures = new Regex(...)` seems more intuitive

luciannojunior:
Nice functionality!

react/components/Image.js
brunojdo: Add TODO, Technical Debt

react/components/Image.js
```diff
@@ -1,15 +1,39 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
+import { withRuntimeContext } from 'render'
+
+const MOBILE_SIZE = 200
```
brunojdo: In my desktop the size is 273. And the mobile is 273 too. I think we can only one type like 400. We need to test more how this scale

PR #70, opened 2018-12-05
Fix unnecessary BuyButton rendering

matheusps:
#### What is the purpose of this pull request? What problem is this solving?
As the title says.

[Workspace](https://fixbuybutton--storecomponents.myvtex.com/)

#### Types of changes

### DIFF
--- a/react/index.js
+++ b/react/index.js
@@ -286,7 +286,31 @@ class ProductSummary extends Component {
       'bb b--muted-4 ma2': showBorders
     })
 
-    const style = { visibility: showBuyButton ? 'visible' : 'hidden' }
+    const buyButtonStyle = { visibility: showBuyButton ? 'visible' : 'hidden' }
+
+    const renderBuyButton = () => (
+      !hideBuyButton && (
+        <div className={buyButtonClasses}>
+          <div style={ buyButtonStyle } className="vtex-product-summary__buy-button center mw-100">
+            <BuyButton
+              available={isAvailable}
+              skuItems={
+                path(['sku', 'itemId'], product) && [
+                  {
+                    skuId: path(['sku', 'itemId'], product),
+                    quantity: 1,
+                    seller: path(['sku', 'seller', 'sellerId'], product),
+                  },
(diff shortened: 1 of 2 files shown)


react/index.js
```diff
@@ -286,7 +286,31 @@ class ProductSummary extends Component {
       'bb b--muted-4 ma2': showBorders
     })
 
-    const style = { visibility: showBuyButton ? 'visible' : 'hidden' }
+    const buyButtonStyle = { visibility: showBuyButton ? 'visible' : 'hidden' }
+
+    const renderBuyButton = () => (
+      !hideBuyButton && (
+        <div className={buyButtonClasses}>
+          <div style={ buyButtonStyle } className="vtex-product-summary__buy-button center mw-100">
```
gugabribeiro: Remove extra spaces around ```buyButtonStyle```.

react/index.js
```diff
@@ -286,7 +286,31 @@ class ProductSummary extends Component {
       'bb b--muted-4 ma2': showBorders
     })
 
-    const style = { visibility: showBuyButton ? 'visible' : 'hidden' }
+    const buyButtonStyle = { visibility: showBuyButton ? 'visible' : 'hidden' }
```
gugabribeiro: Do not you think that should be better to use tachyons css to hide the component instead of styled fields?

react/index.js
```diff
@@ -286,7 +286,31 @@ class ProductSummary extends Component {
...
+
+    const renderBuyButton = () => (
```
Klynger: I think this function should be a getter just like the other render functions of the component

react/index.js
Klynger: Instead of apllying styles it should be better apply a css `class`

react/index.js
gugabribeiro: This is a render task, do not you think that should be better to extract this to a function?

react/index.js
luciannojunior: `classname` ftw

PR #84, opened 2019-01-09
Refactor buy button behavior and change storefront labels

theomoura:
#### What is the purpose of this pull request?
Change buy button behavior for mobile and web devices. The buy button should appear on mobile devices when 'Always Show' option is selected and disappear on the other options.

#### What problem is this solving?
Buy button was never showing on mobile devices and somo storefront options only works on web devices.

#### How should this be manually tested?
[Access this workspace](https://adminshelffix--storecomponents.myvtex.com/admin/cms/storefront)

#### Screenshots or example usage
clipboard 2019-07-01 at 11 25 45 am 1

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #88, opened 2019-01-14
Fix image size and use Img tag instead of background url

rerissondaniel:

Use resized images in product-summary instead of the original image.

Even though the image needed in this component was small, it always loaded the maximum size of the image. Now it loads with 500px, also making it possible to customize this value through props.

[workspace](https://imageresizeps--storecomponents.myvtex.com/)

#### Types of changes
- [X] Bug fix (a non-breaking change which fixes an issue)

### DIFF
--- a/react/components/Image.js
+++ b/react/components/Image.js
@@ -4,11 +4,24 @@ import PropTypes from 'prop-types'
 import productSummary from '../productSummary.css'
 
 /** Image component with 1:1 aspect ratio */
-export default function Image({ src, className }) {
+export default function Image({ src, className, width = 'auto', height = 'auto'}) {
+  const resizeSrc = src => {
+    //Check of src is already in a size defined by the client
+    const isWithMeasures = new RegExp(/[1-9]+-(:?([1-9]+)|(auto))/);
+    if (isWithMeasures.exec(src)) {
+      return src
+    }
+
+    const imageId = src.slice(src.indexOf("/ids/")).split("/")[2]
+    const withSuffix = `${imageId}-${width}-${height}`
+
+    return src.replace(imageId, withSuffix)
+  }
+
   return (
     <div
       className={`${className} ${productSummary.aspectRatio} w-100`}
-      style={{ backgroundImage: `url(${src})` }}
+      style={{ backgroundImage: `url(${resizeSrc(src)})` }}
     />
   )
 }
@@ -18,4 +31,8 @@ Image.propTypes = {
   src: PropTypes.string.isRequired,
   /** Custom classes */
   className: PropTypes.string,
+  /** Image width */
+  width: PropTypes.number,
+  /** Image height */
+  height: PropTypes.number,
 }
--- a/react/components/Image.js
+++ b/react/components/Image.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import productSummary from '../productSummary.css'
 
 /** Image component with 1:1 aspect ratio */
-export default function Image({ src, className, width = 'auto', height = 'auto'}) {
+export default function Image({ src, className, width, height }) {
   const resizeSrc = src => {
     //Check of src is already in a size defined by the client
     const isWithMeasures = new RegExp(/[1-9]+-(:?([1-9]+)|(auto))/);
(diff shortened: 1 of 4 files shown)


react/components/Image.js
```diff
@@ -4,18 +4,46 @@ import PropTypes from 'prop-types'
 import productSummary from '../productSummary.css'
 
 /** Image component with 1:1 aspect ratio */
-export default function Image({ src, className }) {
+export default function Image({ src, className, width, height }) {
+  const resizeSrc = src => {
```
luciannojunior: Couldn't we reuse [this](https://github.com/vtex-apps/product-details/blob/master/react/utils/generateUrl.js)?

react/components/Image.js
```diff
@@ -4,18 +4,46 @@ import PropTypes from 'prop-types'
...
+    //Check of src is already in a size defined by the client
+    const isWithMeasures = new RegExp(/[1-9]+-(:?([1-9]+)|(auto))/);
+    if (isWithMeasures.exec(src)) {
+      return src
+    }
+
+    const imageId = src.slice(src.indexOf("/ids/")).split("/")[2]
```
luciannojunior: The `[2]` accessor is error prone. Use Ramda's `path` instead

react/components/Image.js
```diff
@@ -4,18 +4,46 @@ import PropTypes from 'prop-types'
...
+    const withSuffix = `${imageId}-${width}-${height}`
+
+    return src.replace(imageId, withSuffix)
+  }
+
   return (
     <div
       className={`${className} ${productSummary.aspectRatio} w-100`}
-      style={{ backgroundImage: `url(${src})` }}
+      style={{ backgroundImage: `url(${resizeSrc(src)})` }}
```
luciannojunior: The PR title says that now the `Img` tag will be used. Wouldn't it change something here?

rerissondaniel: It also says WIP. Sorry for having posted so early in our channel

react/components/Image.js
rerissondaniel: We could, but how?

luciannojunior: You could:

1) import the function exported from `product-details`'s utils.
2) create a common component on `store-components` with the algorithm.

I will check with @brunojdo

rerissondaniel: I think that inserting a dependency of `product-details` only for that is too costly

react/index.js
```diff
@@ -114,7 +122,7 @@ class ProductSummary extends Component {
     } = product
 
     let img = (
-      <Image className={`${productSummary.image}`} alt={name} src={imageUrl} />
+      <Image className={`${productSummary.image}`} alt={name} src={imageUrl} width={imageWidth} height={imageHeight}/>
```
brunojdo: It's not the productName I think that is the skuName

rerissondaniel:
> Fix tests

There's a lot of configuration to the tests to work, since that is done in [this PR](https://github.com/vtex-apps/product-summary/pull/84) I'm gonna wait for it to be accepted

react/components/Image.js
gustavorosolem: This same function is being used on **minicart** too.
https://github.com/vtex-apps/minicart/blob/master/react/utils/urlHelpers.js
This function should be a lib or something that allow using the same code.

react/components/Image.js
```diff
@@ -1,21 +1,49 @@
...
+export default function Image({ src, className, width, height, alt }) {
+  const resizeSrc = src => {
+    //Check of src is already in a size defined by the client
+    const isWithMeasures = new RegExp(/[1-9]+-(:?([1-9]+)|(auto))/);
+    if (isWithMeasures.exec(src)) {
+      return src
+    }
+
+    const imageId = path('2', src.slice(src.indexOf("/ids/")).split("/"))
+    const withSuffix = `${imageId}-${width}-${height}`
```
gustavorosolem: Return the query string used on the new **vtex.file-manager** too.
Example: https://github.com/vtex-apps/product-details/blob/master/react/utils/generateUrl.js#L16

PR #94, opened 2019-01-23
Pass more info to BuyButton

lucasecdb:
Same as the title.

As per the new requirements to the `addToCart` event, these new properties are required on the `BuyButton` component.

[Workspace](https://lucas--storecomponents.myvtex.com/st-tropez-shorts/p).

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [x] New feature (a non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Requires change to documentation, which has been updated accordingly.

### DIFF
--- a/react/components/ProductSummaryBuyButton.js
+++ b/react/components/ProductSummaryBuyButton.js
@@ -46,7 +46,7 @@ const ProductSummaryBuyButton = ({
                 name: product.productName,
                 price: path(['sku', 'seller', 'commertialOffer', 'Price'], product),
                 variant: product.sku.name,
-                brand: product.branc,
+                brand: product.brand,
               },
             ]
           }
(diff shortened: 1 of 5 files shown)


react/index.js
```diff
@@ -147,7 +149,7 @@ class ProductSummary extends Component {
     return img
   }
 
-  get renderImageLoader() {
```
Klynger: Why remove the getters?

lucasecdb: because the getters doesn't make sense for a *function*

react/components/ProductSummaryBuyButton.js
```diff
@@ -43,6 +43,10 @@ const ProductSummaryBuyButton = ({
                 skuId: path(['sku', 'itemId'], product),
                 quantity: 1,
                 seller: path(['sku', 'seller', 'sellerId'], product),
+                name: product.productName,
+                price: path(['sku', 'seller', 'commertialOffer', 'Price'], product),
+                variant: product.sku.name,
+                brand: product.branc,
```
rerissondaniel: shouldn't be product.brand?

lucasecdb: yeah, nice catch :-)

PR #100, opened 2019-01-29
Extract components for a better reuse

GustavoMA:

Extracted the components used in Product Summary to help the next step of the refactoring that will reuse this components for each different displayMode component.

[Here](http://extractcomponentsproductsummary--storecomponents.myvtex.com)

#### Types of changes


react/components/ProductSummaryImage.js
```diff
@@ -0,0 +1,52 @@
...
+  const commertialOffer = path(['sku', 'seller', 'commertialOffer'], product)
+
+  let img = (
+    <Image className={`${productSummary.image}`} alt={name} src={imageUrl} />
+  )
+
+  if (showBadge) {
+    img = (
+      <DiscountBadge
+        listPrice={commertialOffer.ListPrice}
```
jgfidelis: commertialOffer may be null by `path` definition. You may use `pathOr` to fix this, passing {} as a default

react/index.js
```diff
@@ -327,7 +214,7 @@ class ProductSummary extends Component {
             <div className={imageContainerClasses}>
               {path(['sku', 'image', 'imageUrl'], product)
                 ? this.renderImage()
-                : this.renderImageLoader()}
+                : (<ImageLoader />)}
```
jgfidelis: no need for parenthesis :P

react/components/ProductSummaryImage.js
```diff
@@ -0,0 +1,52 @@
...
+    productName: name,
+    sku: {
+      image: { imageUrl },
+    },
+  } = product
+
+  const commertialOffer = path(['sku', 'seller', 'commertialOffer'], product)
+
+  let img = (
+    <Image className={`${productSummary.image}`} alt={name} src={imageUrl} />
```
Klynger: No need of string interporlation here

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Extract internal components to better reuse
```
Klynger:
```suggestion
- Extract internal components to better reuse.
```

luciannojunior:
So many conditional blocks and `let`s in `ProductSummaryImage.js` 😢 

And also, I personally think is semantically confusing with an `Image` component inside the `ProductSumary` app and a `ProductSumaryImage` component, and `ProductSummaryImage` with `showCollections=false & showBadge=false` is equal to the `Image` I think you could add the new functionality in the `Image` component with those props with false as default.

And also, don't forget to add the `propTypes` to the brand new components.

Send it to me later

react/index.js
GustavoMA: Ops, fixed!

react/components/ProductSummaryImage.js
GustavoMA: Nice catch. Fixed.

jgfidelis:
please rebase

react/components/ProductImage.js
```diff
@@ -0,0 +1,68 @@
...
+    productClusters,
+    productName: name,
+    sku: {
+      image: { imageUrl },
+    },
+  } = product
+
+  const commertialOffer = pathOr({}, ['sku', 'seller', 'commertialOffer'], product)
+
+  const withBagde = img => (
```
Klynger: Wouldn't be better that these internal functions be defined outside the `ProductImage` component? You would have to make some changes because each function would have to receive their conditionals, but would make this component much cleaner.

react/components/ProductImage.js
```diff
@@ -0,0 +1,68 @@
...
+        listPrice={commertialOffer.ListPrice}
+        sellingPrice={commertialOffer.Price}
+        label={badgeText}
+      >
+        {img}
+      </DiscountBadge>
+    ) : img
+  )
+
+  const withCollection = img => {
```
Klynger: Same thing that I said above

react/components/ProductImage.js
```diff
@@ -0,0 +1,68 @@
...
+      return (
+        <CollectionBadges collectionBadgesText={collections}>
+          {img}
+        </CollectionBadges>
+      )
+    }
+    return img
+  }
+
+  const img = (
```
Klynger: Same

react/components/ImageLoader.js
```diff
@@ -0,0 +1,18 @@
...
+  <ContentLoader
+    style={{
+      width: '100%',
+      height: '100%',
+    }}
+    width={100}
+    height={56}
+    preserveAspectRatio="xMinYMin meet"
+  >
+    <rect width="100%" height="100%" />
```
JuanAlmeida12: Can you use `className`  here? if you can, you should use `w-100` and `h-100` instead of `width="100%" height="100%"`

react/components/ProductImage.js
GustavoMA: It would make the functions outside `ProductImage` much more complex. They would need not only receive their conditionals but some other props used to create the used components, such as `productClusters` in `withCollection`

Klynger: Yean, but they are already some kinda of components, so it would make sense that they get a bit complex.

PR #101, opened 2019-01-29
Using tag img on Image component

theomoura:
Use the tag `img` instead of `background-img` on Image component.

We should use tag `img` for indexing on SEO.

[Access this workspace](https://theo--storecomponents.myvtex.com/)

#### Types of changes


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Replace `bakground-image` for `img` on Image component.
```
luciannojunior: `background-image`**

react/productSummary.css
```diff
@@ -5,7 +5,12 @@
 .imageContainer {}
 .information {}
 .element {}
-.image {}
+.imageInline {
+  max-height: 7rem;
+}
+.imageNormal {
+  max-height: 281px
```
luciannojunior: Avoid `px`

react/productSummary.css
```diff
@@ -5,7 +5,12 @@
 .imageContainer {}
 .information {}
 .element {}
-.image {}
+.imageInline {
+  max-height: 7rem;
```
luciannojunior: Is this really necessary? Couldn't the boundaries be set in a parent component using Tachyons?

react/components/Image.js
```diff
@@ -4,12 +4,11 @@ import PropTypes from 'prop-types'
...
 /** Image component with 1:1 aspect ratio */
-export default function Image({ src, className }) {
+export default function Image({ src, className, description }) {
   return (
-    <div
-      className={`${className} ${productSummary.aspectRatio} w-100`}
-      style={{ backgroundImage: `url(${src})` }}
-    />
+    <div>
+      <img className={`${className}`} src={src} alt={description} />
```
Klynger: No need for string interporlation here

react/productSummary.css
theomoura: I'm afraid it is. It need to be set directly on the image for IE compatibility.

react/components/Image.js
```diff
@@ -4,12 +4,11 @@ import PropTypes from 'prop-types'
...
 
 /** Image component with 1:1 aspect ratio */
-export default function Image({ src, className }) {
+export default function Image({ src, className, description }) {
   return (
-    <div
-      className={`${className} ${productSummary.aspectRatio} w-100`}
-      style={{ backgroundImage: `url(${src})` }}
-    />
+    <div>
```
Klynger: I think this `div` is not necessary and removing this `div` the component would make any sense too.

react/components/Image.js
```diff
@@ -4,12 +4,11 @@ import PropTypes from 'prop-types'
...
+      <img className={className} src={src} alt={description} />
```
klzns: Why do we still need this component? It's the same as using `` directly

PR #103, opened 2019-01-30
fix price size for inline mode and image container size

jgfidelis:
Correct the proportion between image container and info container, their combined width was 120% and was causing an issue while updating item quantity. (The layout was changing and getting weird).

After correcting this, when the price was getting above 1000, it was breaking layout and not fitting the screen, specially with the big `MX` for the Mexican currency.

After
Mobile:
screen shot 2019-01-30 at 6 17 31 pm
screen shot 2019-01-30 at 6 17 41 pm

Desktop:
screen shot 2019-01-30 at 6 42 50 pm
screen shot 2019-01-30 at 6 42 58 pm

#### Types of changes

### DIFF
--- a/react/components/ProductSummaryPrice.js
+++ b/react/components/ProductSummaryPrice.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { path } from 'ramda'
+import { path, prop } from 'ramda'
 import classNames from 'classnames'
 import { Spinner } from 'vtex.styleguide'
 import { ProductPrice } from 'vtex.store-components'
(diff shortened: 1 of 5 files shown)


react/components/ProductSummaryPrice.js
```diff
@@ -36,6 +36,11 @@ const ProductSummaryPrice = ({
     [`${productSummary.priceContainer} pv5`]: !showBorders,
   })
 
+  const sellingPrice = prop('Price')(commertialOffer)
```
luciannojunior: `const sellingPrice = prop('Price', commertialOffer)` to maintain conciseness with your other changes

react/index.js
```diff
@@ -140,11 +140,11 @@ class ProductSummary extends Component {
 
     const imageContainerClasses = classNames(`${productSummary.imageContainer} db`, {
       'w-100 center': displayMode !== 'inline',
-      'w-40': displayMode === 'inline',
+      'w-30': displayMode === 'inline',
     })
 
     const informationClasses = classNames(`${productSummary.information}`, {
-      'w-80 pb2 pl3 pr3 flex flex-column justify-between': displayMode === 'inline',
+      'w-70 pb2 pl3 pr3 flex flex-column justify-between': displayMode === 'inline',
```
kevinch: Are you using a linter? It would fix this:
```
justify-between': displayMode
```
to this:
```
justify-between' : displayMode
```
It's really a detail but it is an easy feature to use nowadays.

PR #119, opened 2019-02-15
Use the new Apollo Local State API to update the product quantity

estacioneto:
**Related to vtex-apps/minicart#115**

ProductQuantityStepper should use the new Apollo Local State API so the quantity update is faster.

We need to update the ProductQuantityStepper to use a faster API and delete the store-resources dependency.

[Access the workspace](https://offline--storecomponents.myvtex.com/)

#### Types of changes

### DIFF
--- a/react/components/ProductQuantityStepper.js
+++ b/react/components/ProductQuantityStepper.js
@@ -9,27 +9,6 @@ import gql from 'graphql-tag'
 
 import { productShape } from '../utils/propTypes'
 
-export const MINICART_ITEMS_QUERY = gql`
-  query {
-    minicart @client {
-      items {
-        id
-        name
-        imageUrl
-        detailUrl
-        skuName
-        quantity
-        sellingPrice
-        listPrice
-        seller
-        index
-        parentItemIndex
-        parentAssemblyBinding
-      }
-    }
-  }
-`
-
 export const UPDATE_ITEMS_MUTATION = gql`
   mutation updateItems($items: [MinicartItem]) {
     updateItems(items: $items) @client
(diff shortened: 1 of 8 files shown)


react/components/ProductQuantityStepper.js
```diff
@@ -1,64 +1,106 @@
...
+          index: itemIndex,
+          quantity,
+          seller: product.sku.sellerId,
         },
-      })
-      const orderForm = await orderFormContext.refetch()
-      this.checkUpdatedQuantity(orderForm.data.orderForm, itemIndex, quantity)
+      ])
     } catch (err) {
       // gone wrong, rollback to old quantity value
```
luciannojunior: What about this?

estacioneto: Resolved with [this](https://github.com/vtex-apps/minicart/pull/116/commits/11bf5df1aeda1ce6cd0d7a554ecb40e5fe4798ae)

luciannojunior: Deleted after the force push, but I trust you, bro.

estacioneto: Oh... [There it is](https://github.com/vtex-apps/minicart/pull/116/commits/408d41ed9535e6ff088f5d9f6e9f6aff304097b7)

react/components/ProductQuantityStepper.js
```diff
@@ -79,4 +121,21 @@ class ProductQuantityStepper extends Component {
   }
 }
 
-export default injectIntl(withToast(orderFormConsumer(ProductQuantityStepper)))
+const withLinkStateItemsQuery = graphql(MINICART_ITEMS_QUERY, {
```
jgfidelis: This should not happen, minicart items query should be done only on minicart.

react/components/ProductQuantityStepper.js
```diff
@@ -1,64 +1,106 @@
...
-      await orderFormContext.updateOrderForm({
-        variables: {
-          orderFormId: orderFormContext.orderForm.orderFormId,
-          items: [{
-            index: itemIndex,
-            id: product.sku.itemId,
-            quantity,
-            seller: product.sku.sellerId,
-          }],
+      await updateItems([
```
jgfidelis: here we should pass only the item being updated

PR #165, opened 2019-06-17
Add prop to show loading state

lucasecdb:
Add a `isPartial` prop to show a loading state on the product, for when it's only locally persisted in the minicart.

Before, we had no visual indication that the item was only locally persisted or not.

[workspace](https://lucas2--storecomponents.myvtex.com/).

#### Types of changes


estacioneto:
Will the image have this `muted` state? I think it can be valuable too...

PR #172, opened 2019-07-17
Use local items update mutation

lucasecdb:
Update the logic to update the items quantity to use a different mutation when the product is a local item only, i.e. doesn't have a `cartIndex` property yet.

This is needed so we can avoid mixing logic from updating items that need to be synchronized with the server and local items only.

The scenario would be to update items that were added when the user is offline and were never synchronized with the server before, i.e. the user didn't return online with the item in the cart.

[workspace](https://lucas2--storecomponents.myvtex.com/), add an item when you're offline and update it's quantity.

#### Types of changes

### DIFF
--- a/react/legacy/components/ProductQuantityStepper.js
+++ b/react/legacy/components/ProductQuantityStepper.js
@@ -16,6 +16,12 @@ export const UPDATE_ITEMS_MUTATION = gql`
   }
 `
 
+export const UPDATE_LOCAL_ITEMS_MUTATION = gql`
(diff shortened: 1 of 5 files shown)


react/legacy/components/ProductQuantityStepper.js
```diff
@@ -16,6 +16,12 @@ export const UPDATE_ITEMS_MUTATION = gql`
   }
 `
 
+export const UPDATE_LOCAL_ITEMS_MUTATION = gql`
```
jgfidelis: create a file with this

PR #190, opened 2019-10-07
Feature/image size

lbebber:
Adds `width` and `height` props to ProductImage. 

Also adds css handles (to ProductImage).

https://lbebber--storecomponents.myvtex.com/
Look at the first shelf, and confirm that the size is set as follows:
```
  "product-summary-image": {
    "props": {
      "width": {
        "desktop": "200",
        "mobile": "100"
      },
      "height": {
        "desktop": "500",
        "mobile": "100"
      }
    }
  },
```

(it takes into consideration the device's pixel density, so the image size might be twice or more depending on your screen, but the element itself should conform to these dimensions)

#### Types of changes


react/components/ProductSummaryImage/ProductImage.js
```diff
@@ -193,6 +232,8 @@ ProductImage.propTypes = {
   /** Display mode of the summary */
   displayMode: PropTypes.oneOf(['normal', 'inline']),
   hoverImageLabel: PropTypes.string,
+  width: PropTypes.number,
```
klzns: Add object type as well

PR #192, opened 2019-10-10
Fix buy button adding to minicart when you have more than one SKU

Klynger:
When you click on the `BuyButton` and the product has more than one SKU you should go to the Product page instead of add to the mini-cart

`BuyButton` is adding the first SKU of the product when you click on it and you have more than one SKU

Go to the [workspace](https://summarybuybutton--storecomponents.myvtex.com/classic?_q=classic&map=ft) and click on the `BuyButton`. You should go to the product page instead of adding to the mini-cart.

#### Types of changes
- [X] Bugfix (a non-breaking change which fixes an issue)

### DIFF
--- a/react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
+++ b/react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
@@ -5,6 +5,8 @@ import { withRuntimeContext } from 'vtex.render-runtime'
 import { equals, path } from 'ramda'
 import classNames from 'classnames'
 import { IOMessage } from 'vtex.native-types'
+import { Button } from 'vtex.styleguide'
+import { Link } from 'vtex.render-runtime'
 
 import { useProductSummary } from 'vtex.product-summary-context/ProductSummaryContext'
 import displayButtonTypes, {
@@ -13,6 +15,18 @@ import displayButtonTypes, {
 } from '../../utils/displayButtonTypes'
 
 import productSummary from '../../productSummary.css'
+
+function checkSKUAvailability(seller) {
+  return path(['commertialOffer', 'AvailableQuantity'], seller) > 0
+}
+
+function checkAvailableSKUsInSellers(item) {
+  return item.sellers && item.sellers.filter(checkSKUAvailability).length > 0
+}
+
+function hasOnlyOneSKU(items) {
+  return items.filter(checkAvailableSKUsInSellers).length === 1
+}
 
 const ProductSummaryBuyButton = ({
   displayBuyButton,
@@ -59,18 +73,36 @@ const ProductSummaryBuyButton = ({
     selectedQuantity,
   })
 
+  const { items = [] } = product
+  const shouldBeALink = hasOnlyOneSKU(items)
+
   return (
     showBuyButton && (
       <div className={containerClass}>
         <div className={buyButtonClasses}>
-          <BuyButton
-            customToastURL={customToastURL}
-            available={isAvailable}
-            skuItems={skuItems}
-            isOneClickBuy={isOneClickBuy}
-          >
-            <IOMessage id={buyButtonText} />
-          </BuyButton>
+          {shouldBeALink ? (
+            <Link
+              className="dib"
+              page="store.product"
+              params={{
+                slug: product && product.linkText,
+                id: product && product.productId,
+              }}
+            >
+              <Button>
+                <IOMessage id={buyButtonText} />
+              </Button>
+            </Link>
+          ) : (
+            <BuyButton
+              customToastURL={customToastURL}
+              available={isAvailable}
+              skuItems={skuItems}
+              isOneClickBuy={isOneClickBuy}
+            >
+              <IOMessage id={buyButtonText} />
+            </BuyButton>
+          )}
         </div>
       </div>
     )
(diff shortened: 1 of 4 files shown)


react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -14,6 +16,18 @@ import displayButtonTypes, {
...
+function checkSKUAvailability(seller) {
+  return path(['commertialOffer', 'AvailableQuantity'], seller) > 0
+}
+
+function checkAvailableSKUsInSellers(item) {
+  return item.sellers && item.sellers.filter(checkSKUAvailability).length > 0
+}
+
+function shouldAddToCart(items) {
+  return items.filter(checkAvailableSKUsInSellers).length === 1
```
klzns: I think the logic here is even simpler.

Just `items.length > 1`

Klynger: But if it is bigger than 1 you have more than one SKU, so you can't add a random SKU to the cart

lbebber: does it add to cart if the item has, say, 3 SKUs, but only 1 available? if so, I don't think this behaviour is desired.

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -59,18 +73,36 @@ const ProductSummaryBuyButton = ({
     selectedQuantity,
   })
 
+  const { items = [] } = product
+  const haveMoreSKUs = shouldAddToCart(items)
```
lbebber: `hasMultipleSKUs`

victorhmp:
```suggestion
  const hasMoreSKUs = shouldAddToCart(items)
```

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -59,18 +73,36 @@ const ProductSummaryBuyButton = ({
...
+              params={{
+                slug: product && product.linkText,
+                id: product && product.productId,
+              }}
+            >
+              <Button>
+                <IOMessage id={buyButtonText} />
+              </Button>
+            </Link>
+          )}
```
victorhmp: This logic might be reversed

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -14,6 +16,18 @@ import displayButtonTypes, {
...
 
+function checkSKUAvailability(seller) {
+  return path(['commertialOffer', 'AvailableQuantity'], seller) > 0
+}
+
+function checkAvailableSKUsInSellers(item) {
+  return item.sellers && item.sellers.filter(checkSKUAvailability).length > 0
+}
+
+function hasOneSKU(items) {
```
iaronaraujo:
```suggestion
function hasOnlyOneSKU(items) {
```

The `Only` in the name makes it clearer that this not true when there are more than one SKU

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
Klynger: It does, the function `checkSKUAvailability` does this check.

lbebber: (talking about this behavior over DM)

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -14,6 +16,10 @@ import displayButtonTypes, {
 
 import productSummary from '../../productSummary.css'
 
+const ALWAYS_GO_TO_PRODUCT = 'alwaysGoToProduct'
+const ADD_TO_CART = 'addToCart'
+const BUY_BUTTOM_BEHAVIOR_OPTIONS = [ALWAYS_GO_TO_PRODUCT, ADD_TO_CART]
```
victorhmp:
```suggestion
const BUY_BUTTON_BEHAVIOR_OPTIONS = [ALWAYS_GO_TO_PRODUCT, ADD_TO_CART]
```

react/components/ProductSummaryBuyButton/ProductSummaryBuyButton.js
```diff
@@ -14,6 +16,10 @@ import displayButtonTypes, {
 
 import productSummary from '../../productSummary.css'
 
+const ALWAYS_GO_TO_PRODUCT = 'alwaysGoToProduct'
+const ADD_TO_CART = 'addToCart'
```
lbebber: (pensando se tem um nome melhor pra esse modo; ele nao vai necessariamente adicionar ao carrinho sempre, ele vai adicionar só se nao tiver variaçoes né?)

lbebber: acho q pode ser só `default` mesmo, e explicar isso na documentaçao 🤷‍♂

PR #193, opened 2019-10-14
Fix/huge placeholder

lbebber:
Fixes issue where the ImagePlaceholder was getting huge and messing up with the layout.

Also makes proptypes slighltly more permissive, to fix warnings

Go to https://www.als.com/clothing/pants/Womens?map=c,c,specificationFilter_7721 and scroll down until about page 6 or 7, and notice that there is a huge margin below the footer.

Doing the same on https://lbebber--alssports.myvtex.com/clothing/pants/Womens?map=c,c,specificationFilter_7721 shouldn't have this issue

#### Types of changes

### DIFF
--- a/react/legacy/components/ProductSummaryPrice.js
+++ b/react/legacy/components/ProductSummaryPrice.js
@@ -68,9 +68,9 @@ ProductSummaryPrice.propTypes = {
   /** Set installments' visibility */
   showInstallments: PropTypes.bool,
   /** Text of selling Price's label */
-  labelSellingPrice: PropTypes.string,
+  labelSellingPrice: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
   /** Text of selling Price's label */
-  labelListPrice: PropTypes.string,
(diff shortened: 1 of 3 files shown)


react/legacy/components/ProductSummaryPrice.js
```diff
@@ -68,9 +68,9 @@ ProductSummaryPrice.propTypes = {
   /** Set installments' visibility */
   showInstallments: PropTypes.bool,
   /** Text of selling Price's label */
-  labelSellingPrice: PropTypes.string,
+  labelSellingPrice: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
   /** Text of selling Price's label */
-  labelListPrice: PropTypes.string,
+  labelListPrice: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
jgfidelis: :wat:

lbebber: ah this is to remove proptype warnings where `labelListPrice` was set to `false` (probably is being set like `labelListPrice={foo && foo.bar}` 

it's weird but 🤷‍♂

PR #197, opened 2019-10-17
Set a max-width for the ProductSummary component

victorhmp:

Limit the max-width of a ProductSummary component.

Prevents `ProductSummary` from being stretched even though the content would not fill it up. 

https://productsummarywidth--storecomponents.myvtex.com/apparel---accessories
(Apply some filters to see what it looks like with less items)

#### Types of changes

### DIFF
--- a/react/components/ProductSummary.js
+++ b/react/components/ProductSummary.js
@@ -12,6 +12,8 @@ import productSummary from '../productSummary.css'
 import { productShape } from '../utils/propTypes'
 import { mapCatalogProductToProductSummary } from '../utils/normalize'
 
+const PRODUCT_SUMMARY_MAX_WIDTH = 300
+
 const ProductSummaryCustom = ({ product, actionOnClick, children }) => {
   const { isLoading, isHovering } = useProductSummary()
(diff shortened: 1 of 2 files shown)


victorhmp:
I did not make this a prop just because it could lead to bad UX. The max-width for the content inside of a `ProductSummary` for some reason is always ~280px. I also don't think it would be a good idea to make it smaller for some reason since you can always control the number of `product-summary` components you can show

react/components/ProductSummary.js
```diff
@@ -75,6 +77,7 @@ const ProductSummaryCustom = ({ product, actionOnClick, children }) => {
         className={containerClasses}
         onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
+        style={{ maxWidth: `${PRODUCT_SUMMARY_MAX_WIDTH}px` }}
```
Klynger: This would work too
```suggestion
        style={{ maxWidth: PRODUCT_SUMMARY_MAX_WIDTH }}
```

PR #198, opened 2019-10-17
Create ProductSummaryList

victorhmp:

Create `ProductSummaryList` component that is capable of performing a GraphQL query for a list of products, create a list of `ProductSummary`s and expose it via context provider.

This enables `vtex.list-context` to export a `list-context.product-list` interface that uses this component.

The first slider in this page:
https://listcontexttest--storecomponents.myvtex.com/about-us

#### Types of changes


react/ProductSummaryList.js
```diff
@@ -0,0 +1,189 @@
...
+      specificationFilters: specificationFilters.map(parseFilters),
+      orderBy,
+      from: 0,
+      to: maxItems - 1,
+      hideUnavailableItems,
+    },
+  }),
+}
+
+function getOrdenationNames() {
```
iaronaraujo:
```suggestion
function getOrdinationNames() {
```
I know in most of the places in our code it is "Ordenation" but this is actually grammatically wrong. The correct term is "Ordination" so I think it is valid to make that change where we can without breaking someone else's code

react/ProductSummaryList.js
```diff
@@ -0,0 +1,189 @@
...
+  const names = []
+  for (const key in ORDER_BY_OPTIONS) {
+    names.push(ORDER_BY_OPTIONS[key].name)
+  }
+  return names
+}
+
+function getOrdenationValues() {
```
iaronaraujo: same

victorhmp:
Still need to figure out how to enable the user to customize `ProductSummary` inside of this `ProductSummaryList`

react/ProductSummaryList.js
```diff
@@ -0,0 +1,193 @@
...
+
+const ProductSummaryList = ({ children, data, intl, buyButtonText }) => {
+  const formattedBuyButtonText = formatIOMessage({ id: buyButtonText, intl })
+  const componentList = data.products.map(product => {
+    const normalizedProduct = mapCatalogProductToProductSummary(product)
+
+    return (
+      <ProductSummary key={product.productId} product={normalizedProduct}>
+        <ProductSummaryImage />
+        <ProductSummaryName />
```
klzns: Can't this point to an extension point so this layout may be flexible?

victorhmp: It did work the way I tried to! I'm not sure how to make this work with `<ExtensionPoint />`, because if I just pass the id `"product-summary"`, or `product-summary.shelf`, for example, it renders nothing, even when these blocks are defined in the store's theme :(
It would surely be ideal!

klzns: We can't go back to the old ways of fixed layout. We need to figure this out.

react/ProductSummaryList.js
```diff
@@ -0,0 +1,193 @@
...
+    },
+  }),
+}
+
+function getOrdinationNames() {
+  const names = []
+  for (const key in ORDER_BY_OPTIONS) {
+    names.push(ORDER_BY_OPTIONS[key].name)
+  }
+  return names
```
Klynger: Maybe you could make a more abstract function like
```suggestion
function getOrdinationProp(propName) {
  return Object.keys(ORDER_BY_OPTIONS).map(key => ORDER_BY_OPTIONS[key][propName])
}
```

or you could generate both functions (`getOrdinationNames` and `getOrdinationValues`)

```javascript
function getOrdinationPropFn(propName) {
  return () => Object.keys(ORDER_BY_OPTIONS).map(key => ORDER_BY_OPTIONS[key][propName])
}

const getOrdinationNames = getOrdinationPropFn('name')
const getOrdinationValues = getOrdinationPropFn('value')
```

react/graphql/products.graphql
```diff
@@ -0,0 +1,79 @@
...
+    brand
+    brandId
+    specificationGroups {
+      name
+      specifications {
+        name
+        values
+      }
+    }
+    items {
```
Klynger: You're missing `variations` from `items`. I don't if that's the case, but shouldn't this be inside [store-resource](https://github.com/vtex-apps/store-resources)?

klzns: I think we should use the one already in store-resource

victorhmp: Awesome! Did not think about that hahaha makes perfect sense

react/ProductSummaryList.js
victorhmp: Nice! Much cleaner

react/graphql/products.graphql
victorhmp: There are some arguments that I need to use in order to perform this query using the one from `store-resources`, I'll add them there

victorhmp: Solved here: https://github.com/vtex-apps/store-resources/pull/68

victorhmp:
Closing this for now.
This approach is not the best in terms of usability, this should be redesigned when we kill `composition` on Store Framework

PR #200, opened 2019-10-21
Fix pixel event when updating item quantity

klzns:
Depends on https://github.com/vtex-apps/minicart/pull/181


react/legacy/components/ProductQuantityStepper.js
```diff
@@ -43,11 +43,14 @@ class ProductQuantityStepper extends Component {
   pushPixelCartEvents = ({ isAdditionOfProd, product }) => {
     // TODO: update these events
```
kevinch: Is that an issue?

klzns: Ops, not anymore

PR #220, opened 2019-12-26
Add inViewRef prop

iaronaraujo:
Add `inViewRef` prop so it would be possible to know when the productSummary is visible.

[Workspace](https://iaronaraujo--storecomponents.myvtex.com/)
Check that the console prints some stuff when the product-summary is 75% visible.

image

#### Types of changes

### DIFF
--- a/react/components/ProductSummary.js
+++ b/react/components/ProductSummary.js
@@ -17,7 +17,7 @@ import { useCssHandles } from 'vtex.css-handles'
 const PRODUCT_SUMMARY_MAX_WIDTH = 300
 const CSS_HANDLES = ['container', 'containerNormal', 'element', 'clearLink']
 
-const ProductSummaryCustom = ({ product, actionOnClick, children }) => {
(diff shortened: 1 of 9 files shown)


react/components/ProductSummary.js
```diff
@@ -17,7 +17,7 @@ import { useCssHandles } from 'vtex.css-handles'
 const PRODUCT_SUMMARY_MAX_WIDTH = 300
 const CSS_HANDLES = ['container', 'containerNormal', 'element', 'clearLink']
 
-const ProductSummaryCustom = ({ product, actionOnClick, children }) => {
```
lbebber: por que especificamente "`inViewRef`"? pq nao só `ref`?

iaronaraujo: For some reason, it doesn't work

lbebber: i see that our webframework doesn't seem to support `forwardRef`. 

still, `inViewRef` is too specific, I think. something like `containerRef` would be more appropriate, I believe

PR #241, opened 2020-02-28
Fix props passed to ProductSummaryList

klzns:
It was not possible to change the query being made by ProductSummaryList. This happened because there was a component in the tree that was not passing the props down to the component that did the query.

You can see here that I've change the query to maxItems 50 and other parameters:
https://breno--storecomponents.myvtex.com/

#### Types of changes


react/ProductSummaryList.js
```diff
@@ -115,12 +117,12 @@ const ProductSummaryList = ({
...
 }
 
-const EnhancedProductList = ({ children }) => {
+const EnhancedProductList = ({ children, ...props }) => {
   const { ProductListProvider } = ProductListContext
 
   return (
     <ProductListProvider>
-      <ProductSummaryList>{children}</ProductSummaryList>
+      <ProductSummaryList {...props}>{children}</ProductSummaryList>
```
klzns: I know I could pass children as prop, but I think that's more clear.

klzns:
Moved to https://github.com/vtex-apps/product-summary/pull/242

PR #246, opened 2020-03-11
Remove graphql tag

tlgimenes:
graphql-tag removed from deps since `parse`, from graphql-js already does the job. It is not ideal to use gql from 'graphql-tag' since our current webpack config only strips away the gql when explicitly importing, and not using gql in the code. 

#### Types of changes

### DIFF
--- a/react/legacy/components/ProductQuantityStepper.js
+++ b/react/legacy/components/ProductQuantityStepper.js
@@ -151,21 +151,15 @@ class ProductQuantityStepper extends Component {
 }
 
 const withUpdateItemsMutation = graphql(UPDATE_ITEMS_MUTATION, {
-  props: ({ mutate }) => {
-    debugger
-    console.log('MUTATION withUpdateItemsMutation')
-    return ({
+  props: ({ mutate }) => ({
     updateItems: items => mutate({ variables: { items } }),
-  })},
+  }),
 })
 
 const withUpdateLocalItemsMutation = graphql(UPDATE_LOCAL_ITEMS_MUTATION, {
-  props: ({ mutate }) => {
-    debugger
-    console.log('MUTATION withUpdateLocalItemsMutation')
-    return ({
+  props: ({ mutate }) => ({
     updateLocalItems: items => mutate({ variables: { items } }),
-  })},
+  }),
 })
 
 export default compose(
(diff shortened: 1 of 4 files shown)


react/legacy/components/ProductQuantityStepper.js
```diff
@@ -6,21 +6,21 @@ import { Pixel } from 'vtex.pixel-manager/PixelContext'
...
-`
+`)
 
-const UPDATE_LOCAL_ITEMS_MUTATION = gql`
+const UPDATE_LOCAL_ITEMS_MUTATION = parse(`
   mutation updateLocalItems($items: [MinicartItem]) {
     updateLocalItems(items: $items) @client
   }
-`
+`)
```
klzns: Isnt it better to create a .graphql file instead?

react/legacy/components/ProductQuantityStepper.js
```diff
@@ -151,15 +151,21 @@ class ProductQuantityStepper extends Component {
 }
 
 const withUpdateItemsMutation = graphql(UPDATE_ITEMS_MUTATION, {
-  props: ({ mutate }) => ({
+  props: ({ mutate }) => {
+    debugger
+    console.log('MUTATION withUpdateItemsMutation')
```
klzns:
```suggestion
```

react/legacy/components/ProductQuantityStepper.js
```diff
@@ -151,15 +151,21 @@ class ProductQuantityStepper extends Component {
...
     updateItems: items => mutate({ variables: { items } }),
-  }),
+  })},
 })
 
 const withUpdateLocalItemsMutation = graphql(UPDATE_LOCAL_ITEMS_MUTATION, {
-  props: ({ mutate }) => ({
+  props: ({ mutate }) => {
+    debugger
+    console.log('MUTATION withUpdateLocalItemsMutation')
```
klzns:
```suggestion
```

react/legacy/components/ProductQuantityStepper.js
tlgimenes: No, this would generate a persisted query. Generating a persisted query means the `__RUNTIME__.cacheHints` will be populated with these two mutations, increasing the size of this object. This would be necessary if these queries get to graphql-server. Since these mutations are client only, we should not import these queries, thus decreasing the overall store size

PR #259, opened 2020-05-13
Fix ProductImage crashing when there is no image to render

Klynger:
As the title says

1. [workspace](https://klynger--storecomponents.myvtex.com/home---decor/?__disableSSR)
2. To test this I removed all images in code to reproduce the use case, but if you want to see a product with an image, just open the quick view of the second item and it will magically display one

image

#### Types of changes
- [X] Bug fix (a non-breaking change which fixes an issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- `ProductImage` crashing if there is no image to display.
 
 ## [2.54.0] - 2020-05-12
 ### Added
-- `href` prop in `ProductSummaryCustom` 
+- `href` prop in `ProductSummaryCustom`.
 
 ## [2.53.3] - 2020-04-24
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,10 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- `ProductImage` crashing if there is no image to render.
 
 ## [2.54.0] - 2020-05-12
 ### Added
-- `href` prop in `ProductSummaryCustom` 
+- `href` prop in `ProductSummaryCustom` .
```
iaronaraujo, line 14:
```suggestion
- `href` prop in `ProductSummaryCustom`.
```

react/components/ProductSummaryImage/ProductImage.js
```diff
@@ -155,19 +155,19 @@ const ProductImageContent = ({
     legacyContainerClasses
   )
 
-  if (!sku || hasError) {
+  const images = pathOr([], ['images'], sku)
```
klzns, line 158:
```suggestion
  const images = sku?.images ?? []
```

react/components/ProductSummaryImage/ProductImage.js
```diff
@@ -155,19 +155,19 @@ const ProductImageContent = ({
...
+  const hoverImage = findImageByLabel(images, hoverImageLabel)
+
+  let imageUrl = pathOr('', ['image', 'imageUrl'], sku)
```
klzns, line 161:
```suggestion
  let imageUrl = sku?.image?.imageUrl ?? ''
```

Klynger: We don't support this features to `js`, so I think we should only try to make this change when we pass these files to `tsx`

Klynger: That's why I'm not removing ramda too

PR #264, opened 2020-06-29
Extract ProductSummaryList without query

thalytafabrine:

We need to extract the `ProductSummaryList` component without the query so that we can use this component to display any product list (for example, products coming from a recommendation API)

#### How to test it?

[Workspace](https://thalyta1--storecomponents.myvtex.com/)

#### Related to / Depends on
https://github.com/vtex-apps/shelf-components/pull/1


store/interfaces.json
```diff
@@ -86,7 +86,7 @@
   "list-context.product-list": {
     "allowed": ["product-summary"],
     "composition": "children",
-    "component": "ProductSummaryList",
+    "component": "ProductSummaryListWithQuery",
```
klzns, line 89: Renaming components is a breaking change.

Even though ProductSummaryList is a good name we should keep it as being used with this component.

```suggestion
    "component": "ProductSummaryListWithQuery",
```

react/components/ProductSummaryList/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <ListContextProvider list={newListContextValue}>
+      {children}
+    </ListContextProvider>
+  )
+}
+
+const ProductSummaryList = ({ children, products }) => {
+  const { ProductListProvider } = ProductListContext
```
klzns, line 39: This could be done outsde of the component scope

```suggestion
  const { ProductListProvider } = ProductListContext
```

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
```
klzns, line 10: Keeping the old name, this PR would be an addition instead of a change

react/ProductSummaryListWithoutQuery.js
```diff
@@ -0,0 +1 @@
+export { default } from './components/ProductSummaryListWithoutQuery'
```
klzns, line 1: Do you mind moving the code to this file instead of creating a new one?

PR #275, opened 2020-09-16
Feature/simulation

hiagolcm:
The search-engines have a performance problem since they need to call the simulation API for each product. This PR allows setting a prop to call this API on the client-side.

I also changed the SVG spinner to a CSS spinner. The SVG one was causing performance problems.

#### How to test it?

[Workspace](https://hiago3--carrefourbrfood.myvtex.com/Alimentos-e-Bebidas/Bebidas?crfint=hm-tlink|alimentos-e-bebidas|1|bebidas|2)

Czffg3e94c

### DIFF
--- a/react/hooks/useIsPriceAsync.js
+++ b/react/hooks/useIsPriceAsync.js
@@ -1,0 +1,11 @@
+import { useSearchPage } from 'vtex.search-page-context/SearchPageContext'
+
+const useIsPriceAsync = () => {
+  const { searchQuery } = useSearchPage()
+
+  return {
+    isPriceAsync: searchQuery?.variables?.simulationBehavior === 'async',
+  }
+}
+
+export default useIsPriceAsync
--- a/react/hooks/useIsPriceAsync.js
+++ b/react/hooks/useIsPriceAsync.js
@@ -1,7 +1,13 @@
 import { useSearchPage } from 'vtex.search-page-context/SearchPageContext'
 
 const useIsPriceAsync = () => {
-  const { searchQuery } = useSearchPage()
+  const searchPage = useSearchPage()
+
+  if (!searchPage) {
+    return false
+  }
+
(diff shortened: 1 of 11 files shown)


iaronaraujo (approved):

react/hooks/useIsPriceAsync.js
```diff
@@ -0,0 +1,11 @@
+import { useSearchPage } from 'vtex.search-page-context/SearchPageContext'
+
+const useIsPriceAsync = () => {
+  const { searchQuery } = useSearchPage()
```
klzns, line 4: Can this destructure break if it's outside of search? I mean, if useSearchPage() returns undefined, what happens?

react/hooks/useSimulation.js
```diff
@@ -0,0 +1,79 @@
...
+    variables: {
+      items: simulationItemsInput,
+    },
+    skip: !isPriceAsync || !inView,
+    ssr: false,
+    onError,
+    onCompleted: (response) => {
+      const simulationItems = response.itemsWithSimulation
+
+      const mergedProduct = clone(product)
```
klzns, line 46: We do we have to merge, can't we just update the whole product data?

klzns (changes requested):

react/hooks/useIsPriceAsync.js
hiagolcm: `useSearchPage` return `{}` when there is no context
anyway, I changed my code to be extra careful

react/hooks/useSimulation.js
hiagolcm: Unfortunately, the simulation API doesn't return the whole item,  so I need to do this merge 😢

thalytafabrine (approved):

klzns (approved):

PR #278, opened 2020-09-25
Add ProductSummary prop to ProductSummaryList

Klynger:

Adding a prop `ProductSummary` to `ProductSummaryList` to be possible to use this component inside of another component'.

I didn't add any documentation about this because it might be confusing of how to use this component since you can pass the block via `blocks`  attribute of the interface

#### How to test it?

[Workspace](https://klynger--marinbrasil.myvtex.com/)

This is my code

```typescript
function ProductSummary(props: SummaryProps) {
  const { product } = props

  return (
    <ProductSummaryCustom product={product}>
      <StackLayout>
        <ProductSummaryImage />
        <ProductSummarySpecificationBadges />
      </StackLayout>
      <ProductSummaryName />
      <RateStars />
      <ProductPrice />
      <AddToCartButton />
    </ProductSummaryCustom>
  )
}

function Shelf(props: ShelfProps) {
  const {
    orderBy,
    category,
    maxItems,
    collection,
    hideUnavailableItems,
    specificationFilters,
  } = props

  return (
    <ProductSummaryList
      orderBy={orderBy}
      maxItems={maxItems}
      category={category}
      collection={collection}
      ProductSummary={ProductSummary}
      hideUnavailableItems={hideUnavailableItems}
      specificationFilters={specificationFilters}
    >
      <SliderLayout />
    </ProductSummaryList>
  )
}
```

#### How does this PR make you feel? [:link:](http://giphy.com/)

Slot because I'm adding a slot... did you get it? 😅

### DIFF
--- a/react/ProductSummaryListWithoutQuery.js
+++ b/react/ProductSummaryListWithoutQuery.js
@@ -8,7 +8,7 @@ import ProductListEventCaller from './components/ProductListEventCaller'
 
 const { ProductListProvider } = ProductListContext
 
-const List = ({ children, products }) => {
+function List({ children, products, ProductSummary }) {
   const { list } = useListContext()
   const { treePath } = useTreePath()
 
@@ -17,6 +17,10 @@ const List = ({ children, products }) => {
       products &&
       products.map((product) => {
         const normalizedProduct = mapCatalogProductToProductSummary(product)
+
+        if (typeof ProductSummary === 'function') {
+          return <ProductSummary product={normalizedProduct} />
+        }
 
         return (
           <ExtensionPoint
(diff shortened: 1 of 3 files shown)


victorhmp (approved):

react/ProductSummaryListWithoutQuery.js
```diff
@@ -18,6 +18,10 @@ const List = ({ children, products }) => {
       products.map((product) => {
         const normalizedProduct = mapCatalogProductToProductSummary(product)
 
+        if (typeof ProductSummary === 'function') {
```
icazevedo, line 21: Just out of curiosity, is it always needed to make this validation when dealing with Slots?

icazevedo (approved):

react/ProductSummaryListWithoutQuery.js
Klynger: @icazevedo I did it just to make sure is what I'm expecting. I think that if someone is using the component right now, and he/she doesn't have the types of the app he/she might think that it should pass an object (like `props.children`) instead of a component, that's why I'm doing this verification. In the end, I'm just trying to be extra safe

PR #284, opened 2020-11-10
Fix orderBy default value

victorhmp:

This enables products to be displayed in the way they were sorted inside a collection. Currently, this is not possible. It also changes the default behavior of the `Relevance` orderBy option, which now will just send an empty string to the API and use the default ordering logic our Seach API uses.

#### How to test it?

[Workspace](https://victormiranda--storecomponents.myvtex.com/)

This is not really a visual change, I don't know how to illustrate it 😢

### DIFF
--- a/react/ProductSummaryList.js
+++ b/react/ProductSummaryList.js
@@ -6,6 +6,10 @@ import { ProductList as ProductListStructuredData } from 'vtex.structured-data'
 import ProductSummaryListWithoutQuery from './ProductSummaryListWithoutQuery'
 
 const ORDER_BY_OPTIONS = {
+  DEFAULT: {
+    name: 'admin/editor.productSummaryList.orderType.default',
+    value: '',
+  },
   RELEVANCE: {
     name: 'admin/editor.productSummaryList.orderType.relevance',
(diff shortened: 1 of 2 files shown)


react/ProductSummaryList.js
```diff
@@ -53,7 +53,7 @@ function ProductSummaryList(props) {
     category = '',
     collection,
     hideUnavailableItems = false,
-    orderBy = ORDER_BY_OPTIONS.TOP_SALE_DESC.value,
+    orderBy = ORDER_BY_OPTIONS.DEFAULT.value,
```
klzns, line 56: Shouldn't it be RELEVANCE?
```suggestion
    orderBy = ORDER_BY_OPTIONS.RELEVANCE.value,
```

klzns (changes requested):

react/ProductSummaryList.js
victorhmp: Yeeees!!

klzns (approved):

kaisermann (approved):

icazevedo (approved):

PR #285, opened 2020-11-13
Adding support for custom placeholder images

icazevedo:

This PR adds support for custom placeholder images on ProductSummaryImage.

#### How to test it?

<s>confia no pai</s> It seems there's no way to create a workspace in which the products are always imageless. I've added some unit test to ease that problem. I've manually disabled product images in the workspace below so it shows a custom placeholder image.

[Workspace](https://icaroimageplaceholder2--storecomponents.myvtex.com/)

This is a custom placeholder image
Screen Shot 2020-11-17 at 12 53 25 PM

#### Related to

[store-components#880](https://github.com/vtex-apps/store-components/pull/880)

### DIFF
--- a/react/components/ProductSummaryImage/ProductImage.js
+++ b/react/components/ProductSummaryImage/ProductImage.js
@@ -138,6 +138,7 @@ const ProductImageContent = ({
   mainImageLabel,
   hoverImageLabel,
   showCollections,
+  placeholder,
   width: widthProp,
   height: heightProp,
   aspectRatio,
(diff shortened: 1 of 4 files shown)


react/components/ProductSummaryImage/ProductImage.js
```diff
@@ -172,7 +173,7 @@ const ProductImageContent = ({
 
   let skuImageUrl = pathOr('', ['image', 'imageUrl'], sku)
 
-  if (!skuImageUrl || hasError) {
+  if (!placeholder && (!skuImageUrl || hasError)) {
```
victorhmp, line 176: This is more of a suggestion, but since you're going to perform the same check for `(!skuImageUrl || hasError)` a few lines down, you could store this result in a variable, maybe `shouldDisplayPlaceholder`, and then this conditional would also become clearer to read:

```js
if (shouldDisplayPlaceholder && !placeholder) {
```

Then in line 224 it would be:

```js
src={shouldDisplayPlaceholder ? placeholder : skuImageUrl}
```

victorhmp (approved):

kaisermann (approved):
Agreed, but I'd do what @victorhmp suggested

Klynger (approved):

PR #308, opened 2021-05-04
Feature/gtm major compatibility

icazevedo:

This PR makes the Product Summary compatible with the new GTM major and fixes two issues: the seller selection wasn't considering default sellers as the rest of the apps and `actionOnClick`  was not passed to ProductSummary when used as slots.

#### How to test it?

[Workspace](https://icarogtm--storecomponents.myvtex.com/)

#### Related to / Depends on

Depends on [https://github.com/vtex-apps/product-summary-context/pull/18](https://github.com/vtex-apps/product-summary-context/pull/18)


react/package.json
```diff
@@ -26,21 +26,22 @@
...
+    "vtex.product-list-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-list-context@0.3.1/public/@types/vtex.product-list-context",
     "vtex.product-specification-badges": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-specification-badges@0.3.0/public/@types/vtex.product-specification-badges",
-    "vtex.product-summary-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary-context@0.7.0/public/@types/vtex.product-summary-context",
-    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.126.4/public/@types/vtex.render-runtime",
-    "vtex.responsive-values": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.responsive-values@0.4.0/public/@types/vtex.responsive-values",
+    "vtex.product-summary": "https://icarogtm--storecomponents.myvtex.com/_v/private/typings/linked/v1/vtex.product-summary@2.71.2+build1620068186/public/@types/vtex.product-summary",
+    "vtex.product-summary-context": "https://icarogtm--storecomponents.myvtex.com/_v/private/typings/linked/v1/vtex.product-summary-context@0.7.0+build1620067144/public/@types/vtex.product-summary-context",
```
icazevedo, line 37: Change this before publishing

react/package.json
```diff
@@ -26,21 +26,22 @@
...
+    "vtex.product-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-context@0.9.9/public/@types/vtex.product-context",
+    "vtex.product-list-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-list-context@0.3.1/public/@types/vtex.product-list-context",
     "vtex.product-specification-badges": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-specification-badges@0.3.0/public/@types/vtex.product-specification-badges",
-    "vtex.product-summary-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary-context@0.7.0/public/@types/vtex.product-summary-context",
-    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.126.4/public/@types/vtex.render-runtime",
-    "vtex.responsive-values": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.responsive-values@0.4.0/public/@types/vtex.responsive-values",
+    "vtex.product-summary": "https://icarogtm--storecomponents.myvtex.com/_v/private/typings/linked/v1/vtex.product-summary@2.71.2+build1620068186/public/@types/vtex.product-summary",
```
icazevedo, line 36: Change this before publishing

react/ProductSummaryList.tsx
```diff
@@ -143,17 +143,19 @@ function ProductSummaryList(props: PropsWithChildren<Props>) {
   })
 
   const { products } = data ?? {}
+  // Not using ?? operator because listName can be ''
+  // eslint-disable-next-line no-unneeded-ternary
+  const listName = rawListName ? rawListName : 'List of products'
```
igorbrasileiro, line 148: You can use OR operator (`||`) for this.

react/ProductSummaryList.tsx
```diff
@@ -143,17 +143,19 @@ function ProductSummaryList(props: PropsWithChildren<Props>) {
...
 
   const productClick = useCallback(
-    (product: any) => {
+    (product: any, position: number) => {
```
igorbrasileiro, line 151: I'm not sure, but if there is something new to add in the future to this function will need to add a 3rd parameter, what do you think about using an object as a second parameter?

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- First available seller being used instead of default seller
```
igorbrasileiro, line 10: I suggest making this phrase more clear to avoid misunderstanding, change the phrase order to something like that:
"Seller to be the default seller instead of the first available seller".

igorbrasileiro: I write this because i understood that should be used the opposite

react/ProductSummarySKUSelector.tsx
```diff
@@ -190,7 +190,7 @@ function ProductSummarySKUSelector(props: Props) {
       ...selectedItem,
       image: selectedItem.images[0],
       seller:
-        getFirstAvailableSeller(selectedItem.sellers) ??
+        getDefaultSeller(selectedItem.sellers as ProductTypes.Seller[]) ??
```
igorbrasileiro, line 193: I don't know about these types 😆 , but you can avoid doing this cast, doing something like this on `getDefaultSeller`.

```ts
interface MinimumSellerNeededProperties { // This name is ilustrative
  sellerDefault: <sellerDefaultTypeHere> 
}

export function getDefaultSeller<T extends MinimumSellerNeededProperties>(sellers?: T[]): T {}
```

igorbrasileiro: Every subtype of MinimumSellerNeededProperties can be used as a parameter and the result of this function also will be typed with the subtype 😃 .

igorbrasileiro (changes requested):
I requested changes only because of the changelog. But if you don't agree with it, let me know, because the code lgtm.

react/ProductSummaryList.tsx
klzns: I think ternary is clearer in this case

react/ProductSummaryList.tsx
klzns: Agree on this one because that's a public API that can evolve and we should avoid breaking changes

react/ProductSummarySKUSelector.tsx
klzns: We can remove the `?? selectedItem.sellers[0]` the function already does this

klzns (changes requested):

react/ProductSummaryCustom.tsx
```diff
@@ -186,6 +186,15 @@ interface Props {
...
   priceBehavior?: 'async' | 'default'
+  /**
+   * Name of the list the Product Summary is in. Should be set by a Shelf or the Search Result gallery.
+   */
+  listName?: string
+  /**
+   * Whether the listName should be sent to product pages via querystring
+   * @default true
+   */
+  trackListName?: boolean
```
igorbrasileiro, line 197: It should be on changelog? If yes, add it there.

igorbrasileiro (approved):

camilavcoutinho (approved):

victorhmp (approved):

PR #326, opened 2021-08-05
Update default seller after simulation

thalytafabrine:
When doing a search using a specific zip code, the product appears as unavailable in PLP even though it has stock in a seller.
This happened because the default seller information was not updated after the simulation call, so the default seller was still the default seller of the store, even though this seller didn't have stock for that zip code.
We need to update the default seller after calling the simulation, ensuring that it will be a seller with stock for the specific zip code, as the `product-summary` uses the default seller to display price and availability information.

Depends on:
https://github.com/vtex-apps/store-graphql/pull/553
https://github.com/vtex-apps/store-resources/pull/158

#### How to test it?

[Workspace](https://regionalization--carrefourbr.myvtex.com/busca/galaxy%20tab%20a%20t290?&map=ft)
- Use the zip code 41750190
- The first product must be available for this zip code

Before:
image

image

After:
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Update default seller after the simulation call.
+
 ## [2.74.0] - 2021-06-28
 ### Changed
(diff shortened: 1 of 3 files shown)


chrsmutti (approved):
LGTM!

ullayne02 (approved):

PR #328, opened 2021-08-10
Set a specific SKU in the product context

Git-the-Sanz:

There was no way of specifying which SKU should be the initial selection in product summaries (Both sliders and galleries).
We are now letting the client specify a logic to then use it for the product-context, which in turn will provide the selected item property to it's childrens.

#### How to test it?

[CHEAPEST](https://cheapest--motoroladevelopment.myvtex.com/)
(You will see the cheapest products selected)

[SPECIFICATION](https://specification--motoroladevelopment.myvtex.com/Smartphones/Moto-G-Family?page=1)
(Look for MOTO G7 PLUS, you will see that the SKU No. 35 is selected, thanks to it's product property called `DefaultSKU`)

For i.e:
By passing "SPECIFICATION" the SKU No. 35 was automatically selected in the Gallery

item

#### Describe alternatives you've considered, if any.

We know that the Admin SKU Position exists, but as of today, this is not working (Confirmed by Catalog team)

#### Related to / Depends on

[Search result PR](https://github.com/vtex-apps/search-result/pull/536)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,14 +3,5 @@
   "root": true,
   "env": {
     "node": true
-  },
-  "overrides": [
-    {
-      "files": ["*.js", "*.jsx"],
-      "parser": "babel-eslint",
-      "parserOptions": {
-        "sourceType": "module"
-      }
-    }
-  ]
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Added
+- The ability to dynamically choose the SKU that will be initially selected
 ## [2.74.1] - 2021-08-09
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - The ability to dynamically choose the SKU that will be initially selected
+
 ## [2.74.1] - 2021-08-09
 
 ### Changed
--- a/docs/ProductSummaryList.md
+++ b/docs/ProductSummaryList.md
@@ -64,5 +64,4 @@ For `PreferredSKUEnum`:
 | Cheapest        | `PRICE_ASC`       | Selects the cheapest SKU in stock it finds.        |
 | Most Expensive  | `PRICE_DESC`      | Selects the most expensive SKU in stock it finds.  |
 
-⚠️ Keep in mind that the default behaviour for the initially selected SKU, has now changed to a Product (field) specification.  
-If that one doesn't exist, it will use the `preferredSKU` prop. You can read more about it, and how to implement it in [Recipes](https://vtex.io/docs/recipes/all)
+⚠️ There's a way to select which SKU should take preference over this prop. You can create a Product (field) specification and per product assign the value of the desired SKU to be initially selected. Keep in mind that If the specification doesn't exist or if the value is empty, it will use the `preferredSKU` prop as fallback. You can read more about it, and how to implement it in [Recipes](https://vtex.io/docs/recipes/all)
--- a/docs/ProductSummaryList.md
+++ b/docs/ProductSummaryList.md
@@ -38,7 +38,7 @@ This block is used to specify what variation of `product-summary` to be used to 
 | `skusFilter`           | `SkusFilterEnum`                       | Control SKUs returned for each product in the query. The less SKUs needed to be returned, the more performant your shelf query will be.                                                                    | `"ALL_AVAILABLE"`        |
 | `installmentCriteria`  | `InstallmentCriteriaEnum`              | Control what price to be shown when price has different installments options.                                                                                                                              | `"MAX_WITHOUT_INTEREST"` |
 | `listName`             | `String`                               | Name of the list property on Google Analytics events.                                                                                                                                                      | ``                       |
-| `preferredSKU`         | `PreferredSKUEnum`                     | Controls which SKU will be selected in the summary                                                                                                                                                          | `"FIRST_AVAILABLE"`      |
+| `preferredSKU`         | `PreferredSKUEnum`                     | Controls which SKU will be selected in the summary                                                                                                                                                         | `"FIRST_AVAILABLE"`      |
 
 For `SkusFilterEnum`:
 
@@ -57,10 +57,12 @@ For `InstallmentCriteriaEnum`:
 
 For `PreferredSKUEnum`:
 
-| Name              | Value             | Description                                                                              |
-| ----------------- | ----------------- | ---------------------------------------------------------------------------------------- |
-| First Available   | `FIRST_AVAILABLE` | Selects the first available SKU in stock it finds.                                       |
-| Last Available    | `LAST_AVAILABLE`  | Selects the last available SKU in stock it finds.                                        |
-| Cheapest          | `PRICE_ASC`       | Selects the cheapest SKU in stock it finds.                                              |
-| Most Expensive    | `PRICE_DESC`      | Selects the most expensive SKU in stock it finds.                                        |
-| A specific SKU ID | `SPECIFICATION`   | Selects the desired SKU, see how to do it in [Recipes](https://vtex.io/docs/recipes/all) |
+| Name            | Value             | Description                                        |
+| --------------- | ----------------- | -------------------------------------------------- |
+| First Available | `FIRST_AVAILABLE` | Selects the first available SKU in stock it finds. |
+| Last Available  | `LAST_AVAILABLE`  | Selects the last available SKU in stock it finds.  |
+| Cheapest        | `PRICE_ASC`       | Selects the cheapest SKU in stock it finds.        |
+| Most Expensive  | `PRICE_DESC`      | Selects the most expensive SKU in stock it finds.  |
+
+⚠️ Keep in mind that the default behaviour for the initially selected SKU, has now changed to a Product (field) specification.  
+If that one doesn't exist, it will use the `preferredSKU` prop. You can read more about it, and how to implement it in [Recipes](https://vtex.io/docs/recipes/all)
--- a/docs/ProductSummaryList.md
+++ b/docs/ProductSummaryList.md
@@ -27,29 +27,40 @@ This block is used to specify what variation of `product-summary` to be used to 
 
 `list-context.product-list` is also responsible for performing the GraphQL query that fetches the list of products, so it can receive the following props:
 
-| Prop name              | Type                                   | Description                                                                                                                                                                                                                               | Default value |
-| ---------------------- | -------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
-| `category`             | `String`                               | Category ID of the listed items. For sub-categories, use "/" (e.g. "1/2/3")                                                                                                                                                               | -             |
-| `specificationFilters` | `Array({ id: String, value: String })` | Specification Filters of the listed items.                                                                                                                                                                                                | []            |
-| `collection`           | `String`                               | Filter by collection.                                                                                                                                                                                                                     | -             |
-| `orderBy`              | `Enum`                                 | Ordination type of the items. Possible values: `''`, `OrderByTopSaleDESC`, `OrderByReleaseDateDESC`, `OrderByBestDiscountDESC`, `OrderByPriceDESC`, `OrderByPriceASC`, `OrderByNameASC`, `OrderByNameDESC` | `OrderByTopSaleDESC`          |
-| `hideUnavailableItems` | `Boolean`                              | Hides items that are unavailable.                                                                                                                                                                                                         | `false`       |
-| `maxItems` | `Number`                              | Maximum items to be fetched.                                                                                                                                                                                                         | `10`       |
-| `skusFilter`               | `SkusFilterEnum`                 | Control SKUs returned for each product in the query. The less SKUs needed to be returned, the more performant your shelf query will be.                                                                                                    | `"ALL_AVAILABLE"` |
-| `installmentCriteria`               | `InstallmentCriteriaEnum`                 | Control what price to be shown when price has different installments options.                                                                                                    | `"MAX_WITHOUT_INTEREST"` |
-| `listName`               | `String`                 | Name of the list property on Google Analytics events.                                                                                                    | `` |
+| Prop name              | Type                                   | Description                                                                                                                                                                                                | Default value            |
+| ---------------------- | -------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
+| `category`             | `String`                               | Category ID of the listed items. For sub-categories, use "/" (e.g. "1/2/3")                                                                                                                                | -                        |
+| `specificationFilters` | `Array({ id: String, value: String })` | Specification Filters of the listed items.                                                                                                                                                                 | []                       |
+| `collection`           | `String`                               | Filter by collection.                                                                                                                                                                                      | -                        |
+| `orderBy`              | `Enum`                                 | Ordination type of the items. Possible values: `''`, `OrderByTopSaleDESC`, `OrderByReleaseDateDESC`, `OrderByBestDiscountDESC`, `OrderByPriceDESC`, `OrderByPriceASC`, `OrderByNameASC`, `OrderByNameDESC` | `OrderByTopSaleDESC`     |
+| `hideUnavailableItems` | `Boolean`                              | Hides items that are unavailable.                                                                                                                                                                          | `false`                  |
+| `maxItems`             | `Number`                               | Maximum items to be fetched.                                                                                                                                                                               | `10`                     |
+| `skusFilter`           | `SkusFilterEnum`                       | Control SKUs returned for each product in the query. The less SKUs needed to be returned, the more performant your shelf query will be.                                                                    | `"ALL_AVAILABLE"`        |
+| `installmentCriteria`  | `InstallmentCriteriaEnum`              | Control what price to be shown when price has different installments options.                                                                                                                              | `"MAX_WITHOUT_INTEREST"` |
+| `listName`             | `String`                               | Name of the list property on Google Analytics events.                                                                                                                                                      | ``                       |
+| `preferredSKU`         | `PreferredSKUEnum`                     | Controls which SKU will be selected in the summary                                                                                                                                                          | `"FIRST_AVAILABLE"`      |
 
 For `SkusFilterEnum`:
 
-| Name | Value | Description |
-| ---- | ----- | ----------- |
+| Name            | Value             | Description                                                                                                                                            |
+| --------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | First Available | `FIRST_AVAILABLE` | Most performant, ideal if you do not have a SKU selector in your shelf. Will return only the first available SKU for that product in your shelf query. |
-| All Available | `ALL_AVAILABLE` | A bit better performace, will only return SKUs that are available, ideal if you have a SKU selector but still want a better performance. |
-| All | `ALL` | Returns all SKUs related to that product, least performant option. |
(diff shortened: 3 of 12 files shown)


react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+        getPriceByCondition(acc, condition)
+        ? curr
+        : acc
+    }
+
+    return getPriceByCondition(curr, condition) <
+      getPriceByCondition(acc, condition)
+      ? curr
+      : acc
+  })
```
icazevedo, line 136: I think you could simplify this by doing something like:

```suggestion
    const price = getPriceByCondition(curr, condition)
    const currentBestPrice = getPriceByCondition(acc, condition)
   
    if (condition === 'expensive') {
      return price > currentBestPrice ? curr : acc
    }

    return price < currentBestPrice ? curr : acc
  })
```

react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+ * that will be looking for
+ * @see {@link https://vtex.io/docs/recipes/all} recipe on how to implement this
+ * */
+function getSpecificSKU(items: any[], specifications: any[]) {
+  const defaultSKUspec =
+    specifications.find(
+      (spec: { name: string }) => spec.name === 'DefaultSKUSelected'
+    ) ?? null
+
+  if (!defaultSKUspec) return items[0]
```
icazevedo, line 152: I'm really not sure about this. Imagine the scenario: the client wants some products to follow their default sku specification, and the rest to be ordered following `PRICE_ASC`. It wouldn't be possible with this code, since if products don't have a default SKU specified, the selected SKU will always be the first one, and there'll be no way of changing that.

What I propose is: `getSpecificSKU` or `findSKUBySpecification` (however you name it) should be the default choosing option, and should always apply if such a specification is set for that product. It shouldn't be a choice on the theme. If the client added the specification to the product, that's what they want, right?

That wouldn't change the behavior of current stores, since they don't have any product with that specification set. This way, after not finding any SKU with that specification, the code would use the specified `PreferredSKU` option, which by default would be `FIRST_AVAILABLE`, thus not changing how the store currently works. What do you think?

icazevedo (changes requested):

react/utils/normalize.ts
```diff
@@ -5,6 +5,15 @@ export const DEFAULT_HEIGHT = 'auto'
...
 export const MAX_HEIGHT = 4000
 
+export type PreferenceType =
+  | 'FIRST_AVAILABLE'
+  | 'LAST_AVAILABLE'
+  | 'PRICE_ASC'
+  | 'PRICE_DESC'
+  | 'SPECIFICATION'
+
+type ConditionRule = 'cheapest' | 'expensive'
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+
+  const availableSellers = sellers.filter(
+    ({ commertialOffer }: any) =>
+      commertialOffer.AvailableQuantity != null &&
+      commertialOffer.AvailableQuantity > 0
+  )
+
+  let itemPrice
+
+  if (condition === 'expensive') {
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -77,21 +176,75 @@ function findAvailableProduct(item: {
...
+  preferredSKU: PreferenceType,
+  properties: any
+) {
+  switch (preferredSKU) {
+    default:
+    case 'FIRST_AVAILABLE':
+      return items.find(findAvailableProduct) || items[0]
+
+    case 'LAST_AVAILABLE':
+      return items.reverse().find(findAvailableProduct) || items.reverse()[0]
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -77,21 +176,75 @@ function findAvailableProduct(item: {
...
+
+    case 'PRICE_ASC':
+      return getBestSKUPrice(items, 'cheapest')
+
+    case 'PRICE_DESC':
+      return getBestSKUPrice(items, 'expensive')
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -77,21 +176,75 @@ function findAvailableProduct(item: {
...
+
+    case 'SPECIFICATION':
+      return getSpecificSKU(items, properties)
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+ * @param items an array of SKUs
+ * @param condition the condition enum used for comparison functions
+ * */
+function getBestSKUPrice(items: any[], condition: ConditionRule) {
+  // First, if none or only 1 sku is available, avoid reducing
+  const filteredItems = items.filter(getOnlyAvailable)
+  if (filteredItems.length === 0) return items[0]
+  if (filteredItems.length === 1) return filteredItems[0]
+
+  return filteredItems.reduce((acc: any, curr: any) => {
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+  return seller.commertialOffer.Price
+}
+
+/**
+ * @description
+ * Find the lowest/highest `Price` between sellers
+ * @param item an SKU
+ * @param condition the condition enum used for comparison functions
+ * */
+function getPriceByCondition(item: { sellers: any }, condition: ConditionRule) {
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
```diff
@@ -67,6 +76,96 @@ export function changeImageUrlSize(
...
+  // First, if there's only 1 seller, avoid filtering
+  const { sellers } = item
+
+  if (sellers.length === 1) return sellers[0].commertialOffer.Price
+
+  const availableSellers = sellers.filter(
+    ({ commertialOffer }: any) =>
+      commertialOffer.AvailableQuantity != null &&
```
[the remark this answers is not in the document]
Git-the-Sanz: I added the null check; but also refactored the methods to get seller/s stock. It should now look cleaner.

react/utils/normalize.ts
```diff
@@ -77,21 +176,75 @@ function findAvailableProduct(item: {
   )
 }
 
+/**
+ * @description
+ * Returns only the available SKUs, checks all sellers
```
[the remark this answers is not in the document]
Git-the-Sanz: Added

react/utils/normalize.ts
Git-the-Sanz: A really good idea @icazevedo!
I made it so it first checks if the property exists, and if so attempts to find the desire SKU ID.
If that fails, it will default to the condition supplied.

Git-the-Sanz:
@icazevedo Thank you for your input.
- I typed everything; except the property `normalizedProduct.sku` (I could't manage to do it, the object is particularly weird; and it's used and mutated-around inside product-summary)
- I renamed some functions. They now reflect better what they do.
- I refactored the reduce function. Since the beginning I felt that this approach was barely performant. I now only run the price lookup function once for each SKU, and then just return the best one using the index from max/min.
- Regarding why I'm using .some() I only care if at least 1 seller has stock in a particular Item, if not, it will help us avoid overhead. That's the difference with the other applications.
- Regarding your comment about the Specification. I implemented that one. We first check if the property exists, if it does, we will attempt to retrieve it; if the property is incorrectly valued, we instead attempt to use the switch to find a suitable one.

Thank you for all your inputs!

react/utils/normalize.ts
```diff
@@ -84,14 +229,35 @@ const defaultSeller = { commertialOffer: { Price: 0, ListPrice: 0 } }
...
+ * */
 export function mapCatalogProductToProductSummary(
-  product: any,
+  product: ProductTypes.Product,
+  preferredSKU: PreferenceType = 'FIRST_AVAILABLE',
   imageSize: string | number = 500
 ) {
   if (!product) return null
-  const normalizedProduct = { ...product }
+  const normalizedProduct: any = { ...product }
```
igorbrasileiro, line 243: Just a doubt, why do you need to type as any?

react/utils/normalize.ts
```diff
@@ -84,14 +229,35 @@ const defaultSeller = { commertialOffer: { Price: 0, ListPrice: 0 } }
 const resizeImage = (url: string, imageSize: string | number) =>
   changeImageUrlSize(toHttps(url), imageSize)
 
+/**
+ * @description
+ * This method is responsible of constructing the product type for the product-context.
+ * The property `product.sku` is built by these specifications
+ * */
```
igorbrasileiro, line 236:
```suggestion
 */
```

react/utils/normalize.ts
```diff
@@ -67,14 +77,149 @@ export function changeImageUrlSize(
...
+) {
+  switch (preferredSKU) {
+    default:
+    case 'FIRST_AVAILABLE':
+      return items.find(getAvailableProduct) ?? items[0]
+
+    case 'LAST_AVAILABLE':
+      return (
+        [...items].reverse().find(getAvailableProduct) ??
+        [...items].reverse()[0]
```
igorbrasileiro, line 214: What do you think about these suggestions?
```suggestion
        items[items.length - 1]
```
or 
```ts
case 'LAST_AVAILABLE':
    const reservedItems = [...items].reverse()
    return (reversedItems.find(getAvailableProduct) ?? reversedItems[0])
```

react/utils/normalize.ts
Git-the-Sanz: I'm sending this as any because by typing the product in the function args, and referencing it again in the normalizeProduct, it would trigger the type check.

react/utils/normalize.ts
Git-the-Sanz: Added

react/utils/normalize.ts
igorbrasileiro: Can you share where it will trigger the type check? It's breaking the typescript build? I can try to help you.a

react/utils/normalize.ts
```diff
@@ -67,14 +77,149 @@ export function changeImageUrlSize(
...
+
+/**
+ * @description
+ * Attempts to retrieve the specific SKU using a custom Product (field)
+ * in the Catalog. The specification has the value of the SKU ID
+ * that will be looking for. If not found by any reason, we default
+ * to finding the preferred SKU
+ * @param items an array of SKUs
+ * @param defaultSKUspec the ID used to find the SKU
+ * @param preferenceFallback the logic for finding an SKU if the ID fails
```
igorbrasileiro, line 176: Ain't an English expert, but Grammarly warned about it.
```suggestion
 * Attempts to retrieve the specific SKU using a custom Product (field)
 * in the Catalog. The specification has the value of the SKU ID
 * that will be looking for. If not found for any reason, we default
 * to finding the preferred SKU
 * @param items an array of SKUs
 * @param defaultSKUspec the ID used to find the SKU
 * @param preferenceFallback the logic for finding an SKU if the ID fails
```

react/utils/normalize.ts
```diff
@@ -84,14 +229,35 @@ const defaultSeller = { commertialOffer: { Price: 0, ListPrice: 0 } }
 const resizeImage = (url: string, imageSize: string | number) =>
   changeImageUrlSize(toHttps(url), imageSize)
 
+/**
+ * @description
+ * This method is responsible of constructing the product type for the product-context.
+ * The property `product.sku` is built by these specifications
```
igorbrasileiro, line 235: Grammarly attacks again haha! Feel free to argue against 
```suggestion
 * @description
 * This method is responsible for constructing the product type for the product-context.
 * The property `product.sku` is built by these specifications
```

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Added
+- The ability to dynamically choose the SKU that will be initially selected
```
igorbrasileiro, line 10:
```suggestion
- The ability to dynamically choose the SKU that will be initially selected

```

igorbrasileiro (approved):
Hey @Git-the-Sanz, marvelous documentation! Great job 👏 🚀 .
I left some comments, can you discuss them?
Also, remember to fix the changelog

Git-the-Sanz (commented):
Added all requested changes

Git-the-Sanz (commented):
Ready for review

.eslintrc
```diff
@@ -3,14 +3,5 @@
...
-  "overrides": [
-    {
-      "files": ["*.js", "*.jsx"],
-      "parser": "babel-eslint",
-      "parserOptions": {
-        "sourceType": "module"
-      }
-    }
-  ]
+  }
```
victorhmp, line 6: Why was this change necessary?

docs/ProductSummaryList.md
```diff
@@ -27,29 +27,42 @@ This block is used to specify what variation of `product-summary` to be used to
...
+
+| Name            | Value             | Description                                        |
+| --------------- | ----------------- | -------------------------------------------------- |
+| First Available | `FIRST_AVAILABLE` | Selects the first available SKU in stock it finds. |
+| Last Available  | `LAST_AVAILABLE`  | Selects the last available SKU in stock it finds.  |
+| Cheapest        | `PRICE_ASC`       | Selects the cheapest SKU in stock it finds.        |
+| Most Expensive  | `PRICE_DESC`      | Selects the most expensive SKU in stock it finds.  |
+
+⚠️ Keep in mind that the default behaviour for the initially selected SKU, has now changed to a Product (field) specification.  
+If that one doesn't exist, it will use the `preferredSKU` prop. You can read more about it, and how to implement it in [Recipes](https://vtex.io/docs/recipes/all)
```
victorhmp, line 68: Is there a recipe for this already? If so, update this link to point directly to the recipe, instead of to the recipe list.

Also, I think we're supposed to be linking to developers.vtex.com instead of vtex.io/docs 😄

docs/ProductSummaryList.md
```diff
@@ -27,29 +27,42 @@ This block is used to specify what variation of `product-summary` to be used to
...
+For `PreferredSKUEnum`:
+
+| Name            | Value             | Description                                        |
+| --------------- | ----------------- | -------------------------------------------------- |
+| First Available | `FIRST_AVAILABLE` | Selects the first available SKU in stock it finds. |
+| Last Available  | `LAST_AVAILABLE`  | Selects the last available SKU in stock it finds.  |
+| Cheapest        | `PRICE_ASC`       | Selects the cheapest SKU in stock it finds.        |
+| Most Expensive  | `PRICE_DESC`      | Selects the most expensive SKU in stock it finds.  |
+
+⚠️ Keep in mind that the default behaviour for the initially selected SKU, has now changed to a Product (field) specification.
```
victorhmp, line 67: I don't quite understand what you mean here. The default component behavior has changed?

react/utils/normalize.ts
```diff
@@ -67,14 +77,149 @@ export function changeImageUrlSize(
...
+ * @description
+ * Find the lowest/highest `Price` between sellers
+ * @param item an SKU
+ * @param condition the condition enum used for comparison functions
+ * */
+function getBestPrice(item: ProductTypes.Item, condition: PriceConditionRule) {
+  // If there's only 1 seller, avoid filtering
+  const { sellers } = item
+
+  if (sellers.length === 1) return sellers[0].commertialOffer.Price
```
victorhmp, line 100: Small nitpick, but we prefer always writing `if`s in this form to make them easier to read.

```suggestion
  if (sellers.length === 1) {
    return sellers[0].commertialOffer.Price
  }
```

react/utils/normalize.ts
```diff
@@ -1,10 +1,20 @@
+import type { ProductTypes } from 'vtex.product-context'
```
victorhmp, line 1: This file has become way more complex than before, no problem with that. But could you write some tests for these new functions?

Since they are simple TS functions, you wouldn't need to use anything besides Jest to do so :)

victorhmp (changes requested):

docs/ProductSummaryList.md
Git-the-Sanz: The education team will create the recipe once this is approved.
Regarding the url, I don't see the Recipes subdirectory in developers.vtex.com

docs/ProductSummaryList.md
Git-the-Sanz: Yes, previously the initially selected SKU was the first available, now it's the product specification

react/utils/normalize.ts
Git-the-Sanz: I'm not familiar with it, but I'll see what I can do

react/utils/normalize.ts
Git-the-Sanz: Added

Git-the-Sanz (commented):
Got tests for the 'PRICE_ASC' and for when the 'SPECIFICATION' is in place. Also fixed other test using the mapCatalogProductToProductSummary method

.eslintrc
victorhmp: Yep, that's fine then :)

docs/ProductSummaryList.md
victorhmp: But wouldn't this affect all stores then? If so this would be a breaking change

docs/ProductSummaryList.md
victorhmp: Then I think you could remove this part for now

Git-the-Sanz (commented):
Updated readme

victorhmp (approved):
Awesome! 👏🏻

react/utils/normalize.ts
Git-the-Sanz: If you remove this type assignment, you'll see that on line 278 the property `sku` does not exist on `normalizedProduct` because when doing this: `const normalizedProduct = { ...product }` TS assigns the original type `ProductTypes.Product` to the new Object. 

You could intersect product and SKU, but you'll also need to type a new referenceId and the new images properties, this is way too much effort for so little gain...

Git-the-Sanz (commented):
Awaiting approval

react/utils/normalize.ts
igorbrasileiro: You're right, but have you ever tried to do something like that?

```suggestion
  const normalizedProduct: Product = { ...product }
```
Product comes from [product-summary-context](https://github.com/vtex-apps/product-summary-context/blob/master/react/ProductSummaryTypes.ts#L14)

If it doesn't work, you can try something like that:

```ts
const normalizedProduct: ProductTypes.Product & { sku?: SingleSKU} = { ...product }
```
[SingleSKU](https://github.com/vtex-apps/product-summary-context/blob/master/react/ProductSummaryTypes.ts#L52)

igorbrasileiro: We need to have everything typed, help others developers, and prevents us from miswriting code.

Git-the-Sanz: ex

No because now this doesn't work... really where's the gain of typing something like this?

Git-the-Sanz: The type is only used here, and nowhere else, and If we needed to arrange it to look like the original types, we would need to change the previous functions not assigned to this PR.

Git-the-Sanz: This wasn't even typed before! I'm just getting the flak the because someone else didn't do it.

icazevedo (approved):

igorbrasileiro (approved):

react/utils/normalize.ts
victorhmp: @Git-the-Sanz I'm going to merge this PR now, but just to let you know for the future, this is definitely **not** how engineering at VTEX is treated. With this mindset of "This wasn't even typed before! I'm just getting the flak because someone else didn't do it. This "normalizedProduct" is not of my doing.". If we're making changes to a bit of code that you can improve, by typing, linting, simplifying... whatever it is, we do it. We're all working to make our products better, and we help each other looking over the quality of the code instead of blaming someone else for now having done something before.

PR #333, opened 2021-08-12
Add asyncOnly1P to priceBehavior prop

thalytafabrine:

get the most updated price by simulation only for seller 1, avoiding making many calls to the API at the same time. 
with this, all other sellers will use the price and availability that are returned by the search query, and only seller 1 will have the price being retrieved asynchronously.

#### How to test it?

[Workspace](https://thalyta--carrefourbr.myvtex.com)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "product-summary",
-  "version": "2.74.1",
+  "version": "2.75.0-hkignore.0",
   "title": "Product Summary",
   "description": "Product summary component",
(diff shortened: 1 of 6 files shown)


hiagolcm (approved):

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "product-summary",
-  "version": "2.74.1",
+  "version": "2.75.0-hkignore.1",
```
victorhmp, line 4: Don't forget to update this before merging :)

victorhmp (approved):

PR #348, opened 2022-02-08
Includes props on product summary image

RogerioCrizan:

A customer would like to have photos merged on the mobile device

#### How to test it?

Yes, the change bring two images

[Workspace](https://iomigrationqa--bawclothing.myvtex.com/Masculino)

before:
image

after:
image


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Feature
+
+- Includes prop on product summary image to handle image label render on mobile
```
victorhmp, line 12: Please update the CHANGELOG according to https://github.com/vtex-apps/store-discussion/blob/master/CONTRIBUTING.md.

docs/ProductSummaryImage.md
```diff
@@ -42,6 +42,7 @@
 | `mainImageLabel` | `string \| object` | Matches the value defined in the `imageLabel` field from the admin's Catalog. Once matched, it defines which product image will be the main image displayed in the Product Summary component. | `undefined`|
 | `hoverImageLabel` | `string` | ![https://img.shields.io/badge/-Deprecated-red](https://img.shields.io/badge/-Deprecated-red) Text value that matches the value defined in the `imageLabel` field from the admin's Catalog. Once matched, it defines which product image will be displayed when the user is hovering. If you set a label and no match is found, no image will be displayed during the hover. *Caution*: Use the `hoverImage` prop instead. | `undefined` | 
 | `hoverImage` | `object` | Defines which criteria should be used to define the hover image according to the product images in the admin's Catalog. | `undefined`|
+| `showImgLabelOnMobile` | `boolean` | Whether image label will be displayed on mobile. | `false` |
```
victorhmp, line 45: We tend to stay away from `boolean` props as often as possible to keep our APIs more stable, without adding new props to components every time we want to add new functionality.

Also, I don't think `ImgLabel` is describing what you're doing here... you're basically forcing the image of the product that would appear when a user hovers the mouse on top of it to be rendered on mobile. Just to render two images. But that would make no sense to users. The value passed to the `hoverImage` should not be used with a different purpose then "image to show when onHover is triggered", otherwise the API becomes confusing

victorhmp (changes requested):
I don't feel comfortable with this implementation, and don't think this is something we would like to add into our app.

Also, this looks like a very specific customer need, and it would make our API confusing if we go ahead with the implementation as is. If you just need this for your use case, I would suggest creating a custom React app that solves it and using that instead :)

docs/ProductSummaryImage.md
RogerioCrizan: Hi Victor, how are you?
I can understand you, but we will not keep in touch with the customer and it would be difficult to apply the updates that VTEX releases and with that the product-summary would be outdated. Is there any way to solve our difficulty?

I await your return.

victorhmp: Hey @RogerioCrizan 👋🏻 !

I don't see how you would have problems with outdated versions. You could release a totally new app that only implements a different version of the `product-summary-image` component. That way, you will still be using the `vtex.product-summary` app, apart from the `product-summary-image` block.

PR #359, opened 2022-07-11
Use h3 by default instead of h1

polishq:

Although Google has said that technically there is nothing wrong with having multiple H1 tags on a page, it is still W3C best practice to limit each page to one H1 tag. Having a single H1 tag helps make the page more semantic and accessible (for instance with users of screen readers). This PR changes the default tag for the `ProductSummaryName` to `h3` instead of `h1` (the thought process being that the title of the PLP would be an `h1`, then there may be a content area with an `h2`, then each product's name becomes an `h3`). 

#### How to test it?

Linked here: https://h3test--esika.myvtex.com/maquillaje/c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


danzanzini (approved):
:shipit:

PR #373, opened 2023-04-12
Add shipping options filter to ProductSummaryList

thalytafabrine:
Add `shippingOptions` filter to `ProductSummaryList` when there is a shipping option stored in the session

#### How to test it?

[Workspace](https://shippingoptions--biggy.myvtex.com/?__bindingAddress=biggy.myvtex.com/)

#### Related to / Depends on

Depends on https://github.com/vtex-apps/store-resources/pull/175


react/ProductSummaryList.tsx
```diff
@@ -130,6 +133,22 @@ function ProductSummaryList(props: PropsWithChildren<Props>) {
...
+    async function getShippingFromSession() {
+      const result = await getShippingOptionFromSession()
+
+      if (result && !equals(result, shippingOptions)) {
+        setShippingOptions(result)
+      }
+    }
+
+    getShippingFromSession()
+  }, [getShippingOptionFromSession, shippingOptions])
```
hiagolcm, line 150:
```suggestion
  useEffect(() => {
    async function getShippingFromSession() {
      const result = await getShippingOptionFromSession()

      if (result) {
        setShippingOptions(result)
      }
    }

    getShippingFromSession()
  }, [getShippingOptionFromSession])
```
Wouldn't this achieve the same result?

When `!equals(result, shippingOptions)` is `true` the `shippingOption` will have the value from `result`
When `!equals(result, shippingOptions)` is `false` the `shippingOption` will also have the value from `result`

hiagolcm (approved):

react/ProductSummaryList.tsx
thalytafabrine: this condition is to avoid a loop, since the function keeps updating the shippingOptions even when it's already the same value as result

icazevedo: @thalytafabrine you can use the `setShippingOptions` callback function to do that. This is more performant.

```suggestion
  useEffect(() => {
    async function getShippingFromSession() {
      const result = await getShippingOptionFromSession()

      if (result) {
        setShippingOptions((currentShippingOptions) => equals(currentShippingOptions, result) ? currentShippingOptions : result)
      }
    }

    getShippingFromSession()
  }, [getShippingOptionFromSession])
```

react/hooks/useSession.ts
```diff
@@ -0,0 +1,34 @@
...
+    headers.append('Content-Type', 'application/json')
+
+    const requestOptions: RequestInit = {
+      method: 'GET',
+      headers,
+      redirect: 'follow',
+    }
+
+    const session = await fetch(
+      `${window.location.origin}/api/sessions?items=public.shippingOption`,
```
icazevedo, line 16: I think this makes the same thing

```suggestion
      `/api/sessions?items=public.shippingOption`,
```

icazevedo (approved):
LGTM

PR #380, opened 2023-08-02
feat: Added a prop to set the max description dynamically

arturmagalhaesjr:

It has been added a way to set the max description which is hardcoded by a constant (120).

This way is possible to set a new property on the **Product Description Component** which makes it possible to set it on the theme blocks.

#### How to test it?

https://artur2--beautycounterqa.myvtex.com/

CleanShot 2023-08-02 at 18 05 52@2x

### DIFF
--- a/react/ProductSummaryDescription.tsx
+++ b/react/ProductSummaryDescription.tsx
@@ -11,10 +11,10 @@ const CSS_HANDLES = ['description'] as const
 
 interface Props {
   classes?: CssHandlesTypes.CustomClasses<typeof CSS_HANDLES>
-  maxSizeDescription?: number
+  descriptionMaxSize?: number
 }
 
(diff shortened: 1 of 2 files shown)


react/ProductSummaryDescription.tsx
```diff
@@ -11,9 +11,10 @@ const CSS_HANDLES = ['description'] as const
 
 interface Props {
   classes?: CssHandlesTypes.CustomClasses<typeof CSS_HANDLES>
+  maxSizeDescription?: number
```
mairatma, line 14: Lgtm, I just wanted to suggest renaming this to `descriptionMaxSize` or `maxDescriptionSize`. Otherwise it seems like this is for setting the description of the max size (at least that's how I understand it when reading).

mairatma (approved):
Code lgtm, just a small suggestion inline

Do you know why the CI is failing, though?

PR #382, opened 2023-09-14
add variant to the products query

hiagolcm:
#### How to test it?

[Workspace](Link goes here!)


chrsmutti (approved):

GabEluan:
FYI: The linter Github Action is failing but it is from an older PR that didn't take that into account, and has nothing to do with this changes. We will continue with the merge and deploy of the App.

PR #384, opened 2023-11-09
Feature: ads analytics data-properties

hcaula:

Includes ads data-properties on Product Summary, which enables the Activity Flow script to observe ad elements and collect metrics.

This makes the block `ads-analytics` obsolete in the app `vtex.sponsored-products`. Now, the merchant doesn't have to include the wrapper in their product summary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- When a product is sponsored, adds HTML data properties so that they can be observed by the Activity Flow script.
+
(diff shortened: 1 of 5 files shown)


filipewl (approved):
From the SF side, this looks harmless enough to me! 

Shall we wait for @thalytafabrine's approval as well?

Don't worry about [the failed linter](https://github.com/vtex-apps/product-summary/actions/runs/6815547983/job/18534897386?pr=384). It's been failed since before, and I've doubled-checked no new error was introduced.

hcaula:
@filipewl Thanks for the quick review!

Since it's near the end of the day, I'll most likely deploy this tomorrow morning, giving @thalytafabrine some more time to review. But, this should be harmless anyways.

PR #390, opened 2023-12-15
Feature: supports multiple sponsored badge positions

hcaula:

In some stores, the layout doesn't work so well with the sponsored tag begin on top of the product's name.

So, we're now adding the option `containerTop`, that puts the sponsored tag on top of the product container in an absolute position. The store can then manipulate the finer details in their CSS.

#### How to test it?

[Workspace](https://caula--biggy.myvtex.com/camisa?_q=camisa&map=ft&__bindingAddress=biggy.myvtex.com/)

Screenshot 2023-12-15 at 16 33 20 image

#### Related to / Depends on

Depends on: https://github.com/vtex-apps/product-summary-context/pull/27

### DIFF
--- a/docs/ProductSummaryShelf.md
+++ b/docs/ProductSummaryShelf.md
@@ -47,6 +47,8 @@ The Product Summary Shelf is the main block exported by the [Product Summary app
 | - | - | - | - |
 | `priceBehavior` |  `enum` | Determines whether the component should fetch the most up-to-date price (`async`) or (`default`). Remember to configure the [Search Result](https://vtex.io/docs/components/content-blocks/vtex.search-result@3.79.1/#configuration)'s `simulationBehavior` prop to `skip` and use the Product Price [`product-price-suspense`](https://github.com/vtex-apps/product-price/blob/master/docs/README.md) block to render a loading spinner while the price information is being fetched. | `default` |
 | `trackListName` | `boolean` | Determines whether the component should send the list name to the product page when the product summary is clicked. Disabling it will prevent the `productDetail` GTM event sent on the PDP to identify from which list the user navigated. | `true` |
+| `sponsoredBadgeLabel` | `String` | The text of the "Sponsored" tag, if applicable. | `"store/sponsored-badge.label"`| 
+| `sponsoredBadgePosition` | `enum` | The text of the "Sponsored" tag, if applicable. Possible values are `titleTop`, `containerTopLeft` and `none`. | `"titleTop"`| 
 
 ## Customization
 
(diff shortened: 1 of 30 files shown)


react/utils/shouldShowSponsoredBadge.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  Product,
+  SponsoredBadgePosition,
+} from 'vtex.product-summary-context/react/ProductSummaryTypes'
+
+const shouldShowSponsoredBadge = (
+  product: Product,
+  position: SponsoredBadgePosition,
+  eligiblePositions: SponsoredBadgePosition[]
```
thalytafabrine, line 9: Is there a scenario where we need this to be an array? Considering the places where you used this function, wouldn't it be simpler for this parameter to be a string with the type `SponsoredBadgePosition`?

react/components/SponsoredBadge.tsx
```diff
@@ -0,0 +1,45 @@
...
+  showLabel?: boolean
+}
+
+export const SponsoredBadge = ({
+  label = 'store/sponsoredBadge.title',
+  showLabel = false,
+}: Props) => {
+  const { handles } = useCssHandles(CSS_HANDLES)
+
+  if (!showLabel) return null
```
thalytafabrine, line 26: since this boolean only checks if should show anything, we could remove it and check this before adding the SponsoredBadge to the tree, like this:
```
{showSponsoredBadge && (
  <SponsoredBadge label={sponsoredBadge?.label} />
)}
```

what do you think?

docs/ProductSummaryShelf.md
```diff
@@ -47,6 +47,8 @@ The Product Summary Shelf is the main block exported by the [Product Summary app
 | - | - | - | - |
 | `priceBehavior` |  `enum` | Determines whether the component should fetch the most up-to-date price (`async`) or (`default`). Remember to configure the [Search Result](https://vtex.io/docs/components/content-blocks/vtex.search-result@3.79.1/#configuration)'s `simulationBehavior` prop to `skip` and use the Product Price [`product-price-suspense`](https://github.com/vtex-apps/product-price/blob/master/docs/README.md) block to render a loading spinner while the price information is being fetched. | `default` |
 | `trackListName` | `boolean` | Determines whether the component should send the list name to the product page when the product summary is clicked. Disabling it will prevent the `productDetail` GTM event sent on the PDP to identify from which list the user navigated. | `true` |
+| `sponsoredBadgeLabel` | `String` | The text of the "Sponsored" tag, if applicable. | `"store/sponsored-badge.label"`| 
+| `sponsoredBadgePosition` | `enum` | The text of the "Sponsored" tag, if applicable. Possible values are `titleTop`, `containerTopLeft` and `none`. | `"titleTop"`|
```
thalytafabrine, line 51: you are using the same description for both

react/utils/shouldShowSponsoredBadge.ts
hcaula: You're right. I left it like this because, initially, we had multiple values for the same "macro" position (for instance, `titleBegin` and `titleTop`).

But, since there's no need for that, I'll remove now and, if we add new values, I'll change back.

react/components/SponsoredBadge.tsx
hcaula: I did it like this thinking that, in some cases, we actually want to render the div, but not the label itself (for positioning purposes). So it would be easier to change when that moment comes.

But, since this is not a need here, I'll change to your suggestion. Better to work with the present.

docs/ProductSummaryShelf.md
hcaula: Ooops, I'll change that! Thanks!!

hcaula: Done https://github.com/vtex-apps/product-summary/pull/390/commits/6d3e179ecf5e47ed5d57506fed0559228f6cb47e

react/components/SponsoredBadge.tsx
hcaula: Done https://github.com/vtex-apps/product-summary/pull/390/commits/6a79bd6927e350e0e00a6450e3ce28b4a26dcdea

react/utils/shouldShowSponsoredBadge.ts
hcaula: Done https://github.com/vtex-apps/product-summary/pull/390/commits/8e06157d099d999464884ae31af4e5660d2d1ac3

chrsmutti (approved):
LGTM

thalytafabrine (approved):

PR #395, opened 2024-09-13
Limit fetchpriority="high" to 2 images on desktop for improved LCP performance

gabridevs:

This change optimizes the fetch priority logic for product images by reducing the number of items with `fetchpriority="high"` on desktop from 4 to 3. This adjustment aligns with Google's recommendation for optimal Largest Contentful Paint (LCP) performance, as outlined in [this article](https://web.dev/articles/optimize-lcp?hl=it#optimize_the_priority_the_resource_is_given).

According to Google:

> Even without lazy loading, browsers don't initially load images with the highest priority because they're not render-blocking resources. You can signal to the browser which resources are more important by using the `fetchpriority` attribute on resources that may benefit from higher priority:
>
> ```html
> 
> ```
>
> We recommend setting `fetchpriority="high"` on an `` element if you think it's the page's LCP element. However, setting a high priority on more than one or two images makes prioritization less effective at improving LCP.

By limiting the `fetchpriority="high"` attribute to only the most critical images (3 on desktop), we improve loading efficiency, freeing up bandwidth for other important resources.

#### How to test it?

You can test the change by visiting a product grid in the linked workspace and observing the load times for the images in the first 2 positions. Ensure that only the first 2 images have `fetchpriority="high"` on desktop and the first image on mobile. This can be verified via the browser's DevTools under the "Network" tab.

[Workspace](https://prdev--italiaonlineqa.myvtex.com)

Production:

Screenshot 2024-09-13 alle 15 03 09

Workspace:

Screenshot 2024-09-13 alle 15 00 51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Limited `fetchpriority="high"` to a maximum of 2 images to improve LCP performance in carousels.
+
 ## [2.90.0] - 2024-08-01
 
 ### Added
--- a/react/ProductSummaryImage.tsx
+++ b/react/ProductSummaryImage.tsx
@@ -446,7 +446,7 @@ function ProductImage({
         ? positionNumber === 1
           ? 'high'
           : 'low'
-        : positionNumber < 4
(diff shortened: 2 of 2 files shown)


vsseixaso:
Hi @gabridevs, the workspace linked in the description is broken. Can you fix it?

vsseixaso (approved):

PR #407, opened 2025-08-21
Change default ProductSummaryName element to div

RichardByrneCP:

The default element for the product title in the PLP is a `<h3>`. However, this tag is only used for decorative and not semantic purposes. There is also no `<h2>` element on the page which is not allowed under accessibility guidelines, as there must be a hierarchy of headings on the page that make semantic sense i.e. h1 -> h2 -> h3

This PR changes the element to a `<div>` in order to address this.

#### How to test it?

[Workspace](https://richardaccessibility1--colprofr.myvtex.com/toothpaste)

##### Before

Screenshot 2025-08-21 at 10 45 35

##### After

Screenshot 2025-08-21 at 10 46 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
  ## [Unreleased]
 
+ - Change default ProductSummaryName element to div for accessibility, as h3 element should not be used only for decorative purposes.
+
 ## [2.90.6] - 2025-07-24
 
  ### Fixed 


vmourac-vtex:
Hi @RichardByrneCP, thanks for submitting your PR!
Although we do see the benefit in modifying the default value of the `tag` parameter, and agree that `div` would, in fact, be the correct value for the component, we consider this a risky move because many stores can customize the product title in PLPs by targeting this tag directly in their CSS, e.g.:

```css
.vtex-product-summary-2-x-nameWrapper h3 {
  font-size: 20px;
  text-transform: uppercase;
}
```

Since `h3` is only the default value for the `tag` param in `ProductSummaryName` we would suggest that you modify your implementation passing `div` as the `tag` value

We will wait for further discussions, if you feel like it. If you consider out solution a valid one, you can close the PR

RichardByrneCP:
Hi @vmourac-vtex 

I've made the change locally as suggested, so fixed for our site. 

Feel free to close the PR if VTEX do not want to move forward with it for all customers.

Thanks for your time.

Richard

vsseixaso:
# Accessibility Feature Flag: a11ySemanticHtmlMigration

> **Implementation PR**: [vtex-apps/store#623](https://github.com/vtex-apps/store/pull/623)

## What it is

A feature flag that controls accessibility modifications that introduce breaking changes through HTML structure changes in components.

**Goal**: Allow gradual migration to semantic HTML without breaking existing implementations.

## Configuration

### 1. Manifest.json
```json
{
  "dependencies": {
    "vtex.render-runtime": "8.x"
  }
}
```

### 2. How to enable
  Admin > Apps > My Apps > VTEX Store > Settings > "**Enable accessibility semantic HTML migration**"

## How to use

```tsx
import React from 'react'
import { useRuntime } from 'vtex.render-runtime'

function ProductQuantity() {
  const { getSettings } = useRuntime()
  const settings = getSettings('vtex.store')
  const useSemanticHtml = settings?.advancedSettings?.a11ySemanticHtmlMigration

  return (
    <>
      {useSemanticHtml ? (
        <label htmlFor="quantity-input">Quantity:</label>
      ) : (
        <div>Quantity:</div>
      )}
      <input 
        id={useSemanticHtml ? "quantity-input" : "qty-123"}
        type="number" 
      />
    </>
  )
}
```

## Benefits
- ✅ Better accessibility (screen readers, keyboard navigation)
- ✅ WCAG 2.1 compliance
- ✅ Controlled migration without immediate breaking changes

## ⚠️ Important
- **Default**: Disabled (`false`)
- **Breaking changes**: CSS and JavaScript may need adjustments
- **Always test** in development before enabling in production

vsseixaso (changes requested):
For this PR, the use of the a11ySemanticHtmlMigration flag will be required (more details can be found in the discussions of this PR).

PR #411, opened 2025-10-07
feat: use the new ad event parameter

rodorgas:

We're changing the encoded event parameters from adId to eventParameters, since it's not the ad ID. The data attribute is still using the data-van-aid since that's what the Activity Flow use to track events (to be changed in future).

#### How to test it?

https://adsonboardinglocalhost--biggy.myvtex.com/camisa?__bindingAddress=biggy.myvtex.com/

image

### DIFF
--- a/react/ProductSummaryCustom.tsx
+++ b/react/ProductSummaryCustom.tsx
@@ -16,7 +16,6 @@ import { useIntl } from 'react-intl'
 import LocalProductSummaryContext from './ProductSummaryContext'
 import ProductPriceSimulationWrapper from './components/ProductPriceSimulationWrapper'
 import { SponsoredBadge } from './components/SponsoredBadge'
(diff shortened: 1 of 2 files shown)


react/ProductSummaryCustom.tsx
```diff
@@ -16,7 +16,6 @@ import { useIntl } from 'react-intl'
 import LocalProductSummaryContext from './ProductSummaryContext'
 import ProductPriceSimulationWrapper from './components/ProductPriceSimulationWrapper'
 import { SponsoredBadge } from './components/SponsoredBadge'
-import getAdsDataProperties from './utils/getAdsDataProperties'
```
vsseixaso: `getAdsDataProperties` só era usado aqui. Com a remoção, faz sentido deletar o arquivo `utils/getAdsDataProperties.ts`.

PR #419, opened 2026-04-30
Keep hideUnavailableItems unset when omitted

hiagolcm:

With Delivery Promise (DP) on, Intelligent Search does not put unavailable products at the end of the list when hideUnavailableItems is false—they are mixed in with available ones.

To avoid that unless the merchant explicitly sets hideUnavailableItems to false, we change the default when DP is active: if the argument is omitted, the search-resolver treats it as true (instead of behaving like false as before). When DP is off, omitted still behaves like false.

In this PR we stop forcing hideUnavailableItems to false so the value can stay undefined when nothing was configured, and the resolver can apply the rule above.

#### How to test it?

[Workspace](https://dptest--aramismenswear.myvtex.com/)

Before (unavailable products mixed with available ones when hideUnavailableItems is not set by developer in the time)
image

After:
image


iago1501 (approved):
Once the comment of copilot is fixed in the docs, it's ready to go

ArthurFerrao (approved):

ArthurFerrao (approved):

PR #425, opened 2026-09-01
fix: drop stale priceToken when the simulation overwrites the price

monteirogc:
## Summary

`itemsWithSimulation` can rewrite a seller's `Price` after the search returned it — regionalized stores, B2B price tables. The simulation query does not select `priceToken`, so the signed price survived the merge in `mergeSellers` and kept asserting the indexed price while the shopper saw the simulated one.

With [vcs.checkout#7190](https://github.com/vtex/vcs.checkout/pull/7190) moving the token from "fallback only" to "always use when passed", forwarding a stale token stops being an edge case and becomes a wrong charge: checkout would close the cart at a price the shopper never saw.

This drops `priceToken` from the merged offer whenever the simulated price differs from the indexed one. Offers without a token, and sellers the simulation does not touch, are unaffected.

## Why here

`mergeSellers` is the only place in the storefront where both prices exist at the same time. Downstream — `BuyButton`, `add-to-cart-button`, `order-items` — only the merged offer is visible, so the divergence is no longer detectable there.

## Open question this is deliberately isolated from

Whether the token is signed at index time or at query time by the Intelligent Search API is still unconfirmed ([STR-1233](https://vtex-dev.atlassian.net/browse/STR-1233)). The comparison lives in a single `if`: if it turns out to be index-time, the reference price becomes the claim inside the JWT instead of `indexedOffer.Price`, and only that line changes.

## Related

[STR-1231](https://vtex-dev.atlassian.net/browse/STR-1231), part of [STR-1132](https://vtex-dev.atlassian.net/browse/STR-1132). Sibling PRs: `store-resources#199`, `add-to-cart-button#161`, `store-graphql#699`, `checkout-graphql#219`, `order-items#92`.

Draft: this is the only one independent of the others and safe to merge on its own, but held until the rollout order is settled.

## Test plan

- [x] `yarn test` — 19 passing, including 4 new cases on `mergeSellers`
- [ ] Regionalized account: token present on the shelf, absent after the simulation reprices the item
- [ ] Non-regionalized account: token unchanged end to end

Note: the `tsc` failure on `ProductSummaryList.tsx` (`vtex.structured-data` typings drift) already reproduces on a clean `master`.

Made with [Cursor](https://cursor.com)

[STR-1233]: https://vtex-dev.atlassian.net/browse/STR-1233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mendescamara (approved):

monteirogc:
Parking this one rather than merging it, after the scope of the rollout was narrowed.

The guard here was designed against the proposal of **always** using the signed price. In that world a regionalized or B2B reprice would reach checkout on every sale, so letting a token assert the indexed price would be a standing price-integrity problem. That proposal was split off into its own track, and for the current rollout the token stays what it originally was: a fallback consumed only while Pricing is unavailable. The divergence risk in that narrower scope was reviewed and deliberately accepted.

Under fallback-only semantics this guard is actually counterproductive. The token is only read while Pricing is down, and at that moment the per-item choice is between closing the cart at the indexed price or failing the sale outright. Dropping the token whenever the simulation rewrote the price would strip the fallback from precisely the regionalized stores the initiative exists to protect — trading away resilience to avoid a price difference the decision explicitly accepts.

Leaving the PR open rather than closing it: the change and its tests still hold, and it becomes a prerequisite again the day the always-use flag is picked back up. Worth re-reading then against how the token is signed, since the comparison reference depends on whether the price is signed at index time or query time.

